PR #109, opened 2017-03-03
Update installation instruction

jslee02:
This PR adds the console commands to build chimera.


README.md
```diff
@@ -25,21 +25,33 @@ $ ./chimera -c <yaml_config_file> -o <output_path> my_cpp_header1.h my_cpp_heade
...
-sudo apt-get update
-sudo apt-get install llvm-3.6-dev llvm-3.6-tools libclang-3.6-dev libedit-dev libyaml-cpp-dev
+$ sudo add-apt-repository 'deb http://llvm.org/apt/trusty/ llvm-toolchain-trusty-3.6 main'
+$ wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
+$ sudo apt-get update
+$ sudo apt-get install llvm-3.6-dev llvm-3.6-tools libclang-3.6-dev libedit-dev libyaml-cpp-dev libboost-dev
+$ git clone https://github.com/personalrobotics/chimera.git
+$ cd chimera
+$ mkdir build && cd build
+$ cmake ..
```
mkoval: Nit: I suggest adding a `-DCMAKE_BUILD_TYPE=Release` here, since it's generally practice to specify a build type and we do so for the Mac OS X instructions below.

README.md
```diff
@@ -25,21 +25,33 @@ $ ./chimera -c <yaml_config_file> -o <output_path> my_cpp_header1.h my_cpp_heade
...
+$ make
 ```
 
 **On Mac OS X**
```
mkoval: Nit: Technically, it's now just [Mac OS](http://www.apple.com/macos/sierra/). 😉

README.md
```diff
@@ -25,21 +25,33 @@ $ ./chimera -c <yaml_config_file> -o <output_path> my_cpp_header1.h my_cpp_heade
...
 ```bash
-brew install yaml-cpp --with-static-lib
-PKG_CONFIG_PATH=/usr/local/Cellar/yaml-cpp/0.5.2/lib/pkgconfig cmake -DLLVM_DIR=/usr/local/opt/llvm/share/llvm/cmake ..
+$ brew install boost llvm
+$ brew install yaml-cpp --with-static-lib 
+$ git clone https://github.com/personalrobotics/chimera.git
+$ cd chimera
+$ mkdir build && cd build
+$ PKG_CONFIG_PATH=/usr/local/Cellar/yaml-cpp/0.5.2/lib/pkgconfig cmake -DCMAKE_BUILD_TYPE=Release \
+    -DLLVM_DIR=/usr/local/Cellar/llvm/3.9.1/lib/cmake/llvm/ ..
```
mkoval: I did some poking around in Homebrew and managed to find a way of generating the correct path:
```bash
-DLLVM_DIR=$(brew --prefix llvm)/lib/cmake/llvm
```
I much prefer this to hardcoding a version number, since it will work if the user copies and pastes it.

Also, I seem to have `yaml-cpp` in my default `PKG_CONFIG_PATH`. If the same is true for you, I'd like to remove the explicit definition of `PKG_CONFIG_PATH`. If it is necessary, please replace the hardcoded path with:
```
PKG_CONFIG_PATH=$(brew --prefix yaml-cpp)/lib/pkgconfig
```

README.md
jslee02: You mean `macOS`, right?

README.md
jslee02: Changed the llvm path to use `(brew --prefix llvm)`.

Let me check `PKG_CONFIG_PATH` when I get home, and update it if necessary.

README.md
jslee02: Done

README.md
jslee02: On my laptop, the default `PKG_CONFIG_PATH` is empty. But it seems the Travis test passes without setting `PKG_CONFIG_PATH` to `yaml-cpp` in #125. I guess the `yaml-cpp.pc` is installed at the standard path where `pkg_check_modules()` searches by default. If it's the case, we don't need `PKG_CONFIG_PATH=$(brew --prefix yaml-cpp)/lib/pkgconfig`.

I couldn't find the manual tells what are the search paths of `pkg_check_modules()`, but saw several other installed package use the same pattern such as `/usr/local/opt/<package_name>/lib/pkgconfig/`.

PR #113, opened 2017-03-07
Add option `LINK_LIBRARIES` to `add_chimera_binding()`

jslee02:
Previously, Chimera didn't try to link the placeholder target against the targets of the C++ project being generated the python binding. This made Chimera not to add the include directories and compiler definitions of the targets to the placeholder target, which lead to the build errors.

This PR adds an option `LINK_LIBRARIES` to `add_chimera_binding()` so that Chimera can link to the being passed targets to the placeholder. The option is not required by default so this change shouldn't break the current test.

#107 should be merged before this PR.


psigen:
I'm currently looking into whether it is possible to run a cmake hook that copies the target properties from the virtual target to the placeholder.

I really don't like this mechanism because it requires knowing what LINK_LIBRARIES should be when the target is first declared, which is not required by `target_link_libraries` and might be very inconvenient for selective compilation of certain targets.

jslee02:
Could we merge this as a stopgap until we find a better solution? `dartpy` cannot be built without this. Also, `LINK_LIBRARIES` is an optional argument so wouldn't hurt any existing consumer projects.

mkoval:
I am fine with that. It seems straightforward to maintain backwards compatibility, if necessary to do so, if we find a better alternative. @psigen Do you strongly oppose?

psigen:
Sure, if we can add some comments to make it clear that this is a temporary feature and what we need to remove later.

jslee02:
Comment added!

psigen:
Thanks JS!

We should really make a unit test that checks whether this linking is working.  A simple binding that depends on a library like `-lm` or something just to test this specific issue.  I totally missed this issue before because I was building trivial test bindings that did not use any external libraries.
[This doesn't have to be part of the this PR, just noting it here.]

jslee02:
Seems we can merge this PR!

PR #116, opened 2017-03-10
Add license

jslee02:
This PR adds `LICENSE` file of 3-clause BSD hoping this makes the release process easier. @mkoval or @psigen can freely update or decline this PR if the license type is not the one you want.


mkoval:
This looks good to me. Thanks for taking the time to add this!

@psigen Can you confirm that you're also okay with a three clause BSD license?

psigen:
Sounds good to me.

PR #117, opened 2017-03-11
Rectify unintended conversions of html special entities

jslee02:
I'm not sure if this is an issue of Chimera yet, but some generated bindings contain `&lt;` and `&gt;` that seemed to be converted from `<` and `>`.

This PR adds CMake function `rectify_angle_brackets()` that converts the angle brackets back as a workaround.

This PR contains #113. If #113 is declined, this PR needs to be rebased accordingly.

Edit: Now `&amp;` is replaced to `&`.


psigen:
I think @mkoval and I have localized the underlying issue here to HTML escaping in `mstch`, which we can simply disable by adding the following lambda:

https://github.com/no1msd/mstch#custom-escape-function
```
mstch::config::escape = [](const std::string& str) -> std::string {
  return str;
};
```

We should **not** merge this PR and instead add the escape config to not generate the invalid characters in the first place.

jslee02:
That solution makes much more sense. This PR was only a stopgap. Closing this PR.

Could you point out where we should put that lambda?

jslee02:
TIL: According to the [manual](http://mustache.github.io/mustache.5.html), we could use the triple mustache like `{{{name}}}`. Disabling the HTML escaping would be a simpler solution, though.

PR #123, opened 2017-03-17
Propagate make commands to the child process

jslee02:
This PR addresses #133 by propagating the make commands to the child process.

Reference: http://stackoverflow.com/a/33171336

### DIFF
--- a/cmake/chimeraFunctions.cmake
+++ b/cmake/chimeraFunctions.cmake
@@ -135,7 +135,7 @@ function(add_chimera_binding)
     ExternalProject_Add("${binding_TARGET}_REBUILD"
         DOWNLOAD_COMMAND ""
         INSTALL_COMMAND ""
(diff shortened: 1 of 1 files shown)


cmake/chimeraFunctions.cmake
```diff
@@ -135,7 +135,7 @@ function(add_chimera_binding)
     ExternalProject_Add("${binding_TARGET}_REBUILD"
         DOWNLOAD_COMMAND ""
         INSTALL_COMMAND ""
-        BUILD_COMMAND make "${binding_TARGET}_SOURCES"
+        BUILD_COMMAND $(MAKE) "${binding_TARGET}_SOURCES"
```
mkoval: 😱 I've never seen this feature before. Nice find.

Can you please add the StackOverflow link as a comment in the source code? I am concerned that we someone - possibly myself - will be baffled by this in the future.

jslee02: Done

PR #124, opened 2017-03-17
Support for Clang 3.8 and above

mkoval:
These changes were necessary to get Chimera to build against Clang 4.0.0 on macOS. I believe this should: (1) not break backwards compatibility with Clang 3.6 and (2) should add support for the intermediate versions (i.e. 3.7, 3.8, and 3.9).

This may require some back-and-forth with Travis to make sure that it works on all those versions.


mkoval:
I tested this locally with `brew install llvm`, so I will merge this as it is. I'll address any issues that #125 uncovers in that pull request.

PR #125, opened 2017-03-17
Update Travis for macOS

mkoval:
**Depends On:** #124 

I took a stab at updating Travis to run against the latest version of LLVM on macOS. This will likely require some iteration on Travis until I work the kinks out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,6 @@ matrix:
       env: BUILD_TYPE=Release COVERALLS=OFF COMPILER=CLANG LLVM_VERSION=3.7
     - os: osx
       compiler: clang
-      env: BUILD_TYPE=Debug COVERALLS=OFF COMPILER=CLANG LLVM_VERSION=3.8
-    - os: osx
-      compiler: clang
-      env: BUILD_TYPE=Release COVERALLS=OFF COMPILER=CLANG LLVM_VERSION=3.8
-    - os: osx
-      compiler: clang
       env: BUILD_TYPE=Debug COVERALLS=OFF COMPILER=CLANG LLVM_VERSION=4.0
     - os: osx
       compiler: clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,7 +39,7 @@ script:
   - mkdir build
   - cd build
   - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="/usr/share/llvm-${LLVM_VERSION}/cmake/" ..; fi
-  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="$(brew --prefix llvm@${LLVM_VERSION})/lib/llvm-${LLVM_VERSION}/share/llvm/cmake/" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="$(brew --prefix ${LLVM_PACKAGE})/lib/llvm-${LLVM_VERSION}/share/llvm/cmake/" ..; fi
   - make -j4
   - make test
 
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -8,32 +8,38 @@ matrix:
...
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then . ci/install_linux.sh; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then . ci/install_macos.sh; fi
 
 script:
   - mkdir build
   - cd build
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DLLVM_DIR="/usr/share/llvm-3.6/cmake/" ..; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DLLVM_DIR="/usr/local/Cellar/llvm@3.7/3.7.1/lib/llvm-3.7/share/llvm/cmake/" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="/usr/share/llvm-${LLVM_VERSION}/cmake/" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="$(brew --prefix llvm@${LLVM_VERSION})/lib/llvm-${LLVM_VERSION}/share/llvm/cmake/" ..; fi
```
jslee02: It seems we should use if-statement here as well to get the correct prefix when `LLVM_VERSION=latest`, or let `ci/install_macos.sh` to set a variable according to the LLVM version like `llvm@3.x` (for 3.6, 3.7, 3.8) or `llvm` (for lastest) so that we can use the variable here.

mkoval: Yes, you're right. I missed that. :sweat:

.travis.yml
```diff
@@ -8,32 +8,38 @@ matrix:
...
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then . ci/install_linux.sh; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then . ci/install_macos.sh; fi
 
 script:
   - mkdir build
   - cd build
-  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DLLVM_DIR="/usr/share/llvm-3.6/cmake/" ..; fi
-  - if [ "$TRAVIS_OS_NAME" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DLLVM_DIR="/usr/local/Cellar/llvm@3.7/3.7.1/lib/llvm-3.7/share/llvm/cmake/" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="/usr/share/llvm-${LLVM_VERSION}/cmake/" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then cmake -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DLLVM_DIR="$(brew --prefix ${LLVM_PACKAGE})/lib/llvm-${LLVM_VERSION}/share/llvm/cmake/" ..; fi
```
jslee02: I don't know why but the cmake module directory of the latest llvm is different from other versions. It's:
```
$(brew --prefix ${LLVM_PACKAGE})/lib/cmake/llvm # the latest llvm
$(brew --prefix ${LLVM_PACKAGE})/lib/llvm-${LLVM_VERSION}/share/llvm/cmake/ # other llvm versions
```

We might need to export `LLVM_DIR` in `ci/install_macos.sh` as well depending on the llvm unless we tweak the location to be installed, which might not a good idea.

mkoval:
I think I finally have this working 😓 

Note that I intentionally excluded LLVM 3.8 from testing on macOS because the`llvm@3.8` package is broken: its `llvm-config.cmake` file includes an unexpanded CMake template variable:
```cmake
set(LLVM_DEFINITIONS @LLVM_DEFINITIONS@)
```

@jslee02 Can you do another review pass?

jslee02:
> its llvm-config.cmake file includes an unexpanded CMake template variable

Yeah, I saw that too.

👍 This PR looks good to me! I'm happy to see Chimera is built on macOS with different versions of LLVM toolchains.

PR #128, opened 2017-03-21
Add option whether build mstch or use installed one

jslee02:
I realized that Homebrew contains [mstch package](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mstch.rb) while Ubuntu doesn't, and installing Chimera with the internal mstch conflicts when mstch is installed on macOS.

This PR adds an option whether build the internal mstch or use installed mstch. The option is ON (build internal mstch) so that the default behavior is the same as before. We might want to set the option OFF when building on macOS to avoid the confliction.

The [Homebrew formula for Chimera](https://github.com/personalrobotics/homebrew-tap/blob/master/Formula/chimera.rb) will be updated accordingly once this is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,6 +39,8 @@ script:
   - mkdir build
   - cd build
   - cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" ..
+  - if [ "${TRAVIS_OS_NAME}" = "linux" ]; then cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" "-DCHIMERA_BUILD_MSTCH=ON" ..; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx"   ]; then cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" "-DCHIMERA_BUILD_MSTCH=OFF" ..; fi
   - make -j4
   - make test
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,9 @@ project(chimera)
 cmake_minimum_required(VERSION 2.8)
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake/")
 
+option(CHIMERA_BUILD_MSTCH
+    "Build external/mstch instead of using installed mstch" TRUE)
+
 set(RUNTIME_INSTALL_DIR "bin")
 set(SHARE_INSTALL_DIR "share/${PROJECT_NAME}")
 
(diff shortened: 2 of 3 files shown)


psigen:
Why does the internal mstch conflict?  We are pretty sensitive to the mstch
version.  Perhaps we should switch to compiling it statically into the
executable?

jslee02:
Internal mstch conflicts when we install Chimera using Homebrew. If I has mstch installed and then try to install chimera, I get this error:
```
$ brew install chimera
==> Installing chimera from personalrobotics/tap
==> Downloading https://github.com/personalrobotics/chimera/archive/4199c955c2145e83c
==> Downloading from https://codeload.github.com/personalrobotics/chimera/tar.gz/4199
######################################################################## 100.0%
==> cmake . -DCMAKE_C_FLAGS_RELEASE=-DNDEBUG -DCMAKE_CXX_FLAGS_RELEASE=-DNDEBUG -DCMA
==> make install
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink include/mstch/mstch.hpp
Target /usr/local/include/mstch/mstch.hpp
is a symlink belonging to mstch. You can unlink it:
  brew unlink mstch

To force the link and overwrite all conflicting files:
  brew link --overwrite chimera

To list all files that would be deleted:
  brew link --overwrite --dry-run chimera

Possible conflicting files are:
/usr/local/include/mstch/mstch.hpp -> /usr/local/Cellar/mstch/1.0.2/include/mstch/mstch.hpp
/usr/local/lib/cmake/mstch/mstch-config-version.cmake
/usr/local/lib/cmake/mstch/mstch-config.cmake
/usr/local/lib/cmake/mstch/mstch-targets-release.cmake
/usr/local/lib/cmake/mstch/mstch-targets.cmake
/usr/local/lib/cmake/mstch/mstch-config-version.cmake
/usr/local/lib/cmake/mstch/mstch-config.cmake
/usr/local/lib/cmake/mstch/mstch-targets-release.cmake
/usr/local/lib/cmake/mstch/mstch-targets.cmake
/usr/local/lib/libmstch.a -> /usr/local/Cellar/mstch/1.0.2/lib/libmstch.a
==> Summary
🍺  /usr/local/Cellar/chimera/0.0.1-prerelease: 15 files, 18.4MB, built in 24 seconds
```

psigen:
Hmm, why is this trying to install `include/mstch/mstch.hpp` at all here?
We don't need it for a chimera installation once the executable is
generated.

jslee02:
mstch is built as a static library. The problem is that mstch is installed when Chimera is installed because Chimera includes `external/mstch` directory and just follows the `CMakeLists.txt` of mstch which installs `mstch.hpp` and `libmstch.a`.

Hm, is there a way to prevent installing mstch but being built and statically linked to Chimera?

psigen:
> The problem is that mstch is installed when Chimera is installed because Chimera includes external/mstch directory and just runs the CMakeLists.txt of mstch which installs mstch.hpp and libmstch.a.

I don't think we should be calling _add_subdirectory_ the way that we are, for this reason.  We should probably instead be using an `ExternalProject_Add` and creating `IMPORTED` targets that we can use.

jslee02:
> I don't think we should be calling add_subdirectory() the way that we are, for this reason. We should probably instead be using an ExternalProject_Add and creating IMPORTED targets that we can use.

This sounds more relevant approach than this PR since we don't want to install mstch even on linux. Closing this PR and opening an issue instead.

PR #132, opened 2017-03-28
Added fix for HTML escaping.

### LINKED ISSUE #104: enum ".value"s are being HTML encoded
This may be happening somewhere in the `mstch` templating code. E.g.:

```
.value("Delegate", dart::common::detail::ComposeData&lt;dart::common::detail::CompositeData&lt;std::map&lt;std::type_index, std::unique_ptr&lt;dart::common::Aspect::Properties, std::default_delete&lt;dart::common::Aspect::Properties&gt; &gt;, std::less&lt;std::type_index&gt;, std::allocator&lt;std::pair&lt;const std::type_index, std::unique_ptr&lt;dart::common::Aspect::Properties, std::default_delete&lt;dart::common::Aspect::Properties&gt; &gt; &gt; &gt; &gt;, dart::common::detail::GetProperties&gt;, dart::common::detail::GetProperties, dart::dynamics::BodyNode&gt;::DelegateTag::Delegate)
```

psigen:
This commit adds a custom escaping function to the mstch namespace
that disables all HTML escaping.  Since HTML escaping heavily
interferes with generation of valid C++ type declarations, this
was causing numerous issues in code generation.

As a side effect, the _unescaped_ template include operator for
mstch, `{{{ }}}`, should no longer be necessary, so I have
removed it from all instances of templates in the source tree.

Closes #104.

### DIFF
--- a/include/chimera/mstch.h
+++ b/include/chimera/mstch.h
@@ -31,6 +31,18 @@ namespace mstch
 ::mstch::node wrapYAMLNode(const YAML::Node &node);
 
 /**
+ * Custom escape function that disables HTML escaping on mstch output.
+ *
+ * This is not desirable in chimera because many C++ types include characters
+ * that can be accidentally escaped, such as `<>` and `&`.
+ *
+ * See: https://github.com/no1msd/mstch#custom-escape-function
+ */
+::mstch::config::escape = [](const std::string& str) -> std::string {
+  return str;
+};
+
+/**
  * Base mstch wrapper for Clang declarations.
  */
 template<typename T>
@@ -184,7 +196,7 @@ public:
     ::mstch::node constructors();
     ::mstch::node methods();
     ::mstch::node staticMethods();
-    
+
     ::mstch::node fields();
     ::mstch::node staticFields();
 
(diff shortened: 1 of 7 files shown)


psigen:
I'm not sure if there's a better place to put the escape function directive, but it seemed reasonable to put it into the constructor `CompiledConfiguration` since that is the object that performs all the `mstch::render` calls.

While `CompiledConfiguration` is only ever instantiated once in `chimera`, if we wanted to be super safe we could move the call into a `static const` member variable initialized by a lambda to guarantee that the initialization would only happen once before execution started.

jslee02:
I'm not against with this changes as long as this works for our use, but how about we just use the unescaped template `{{{ }}}` where we need (maybe everywhere). I haven't tried to find the spot, but probably the build error I found was because of the use of the escaped template, which shouldn't be used.

psigen:
Well, in this PR, we force chimera to escape _all_ template variables at _all_ times, so it seems needless to intersperse the `{{ }}`` with `{{{ }}}` which just confuses users with an additional decision when writing templates which as no effect.

I removed all `{{{ }}}` instances, and I think we should specifically not recommend ever using this directive in any documentation.  It is guaranteed that chimera _never_ HTML escapes any template variables under any circumstances.

PR #133, opened 2017-03-28
Converts mstch build to an ExternalProject.

### LINKED ISSUE #129: Don't install mstch when installing Chimera
mstch, which is included in the Chimera source tree, is only necessary for building Chimera as a static library. However, mstch is unnecessarily installed with Chimera because mstch is built using `add_subdirectory()`. This causes confliction when mstch is installed as a separate package using a package manager.

As suggested by @psigen in #128, we should probably instead be using an `ExternalProject_Add` and creating `IMPORTED` targets that we can use.

psigen:
Previously, mstch was built in-source using an `add_subdirectory()` directive.  This caused its `EXPORT` and `INSTALL` directives to be forwarded into chimera's own install step.

This PR changes the build process to use an `ExternalProject` to generate the static `mstch` library, which prevents this forwarding.

Closes #129.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -26,4 +26,3 @@ set_target_properties(mstch PROPERTIES
     INTERFACE_INCLUDE_DIRECTORIES "${mstch_INCLUDE_DIR}"
     IMPORTED_LOCATION "${mstch_LIBRARY}"
 )
-message("JUST ADDED TARGET MSTCH")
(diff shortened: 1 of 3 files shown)


external/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+# Create an IMPORTED target so that CMake can propagate the build properties of
+# this ExternalProject to dependent targets.
+# (Note: This target must be GLOBAL as per: http://stackoverflow.com/a/26315239)
+add_library(mstch STATIC IMPORTED GLOBAL)
+add_dependencies(mstch mstch_EXTERNAL)
+set_target_properties(mstch PROPERTIES
+    INTERFACE_INCLUDE_DIRECTORIES "${mstch_INCLUDE_DIR}"
+    IMPORTED_LOCATION "${mstch_LIBRARY}"
+)
+message("JUST ADDED TARGET MSTCH")
```
mkoval: :rage4:

PR #134, opened 2017-03-28
Staged `sources.txt` generation in CMake binding helpers.

### LINKED ISSUE #114: Don't generate `sources.txt` when binding generation is failed
Chimera generates `sources.txt` even when it fails to generate the bindings (*.cpp files). This makes re-running `make` in the consumer project (e.g., `dartpy`) not to re-run `chimera` to generate the bindings again but just try to build the bindings assuming they are all generated.

The right fix is to modify Chimera to only write `sources.txt` after it is completely done. So `make` will re-run `chimera` to generate the bindings.

The in-the-meantime solution would be deleting `sources.txt` before re-run `make` when the binding generation is failed.

This issue was discovered in the chat with @mkoval .

psigen:
**This PR is included in #135, so optionally we can just review and merge that one.**

Previously, `sources.txt` was directly piped from the output of chimera, but was not cleared or invalidated if the command did not complete successfully, leading to subsequent build steps skipping the regeneration of sources.

This PR changes the `add_custom_command()` directive to stage the output of chimera to a temp file first, which is moved into a final location after the success of the generation command.  This prevents incomplete or invalid `sources.txt` from being used by subsequent build steps.

As a side effect of this PR, I discovered a bug in the unit test runner `CMakeLists.txt` that was leading to the test practically always succeeding because it did not have the `chimeraExecutable` variable set and somehow managed to run a very ill defined bash command that always succeeded.

Closes #114.

### DIFF
--- a/cmake/chimeraFunctions.cmake
+++ b/cmake/chimeraFunctions.cmake
@@ -18,8 +18,8 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 #                     [LINK_LIBRARIES item1 [item2 [...]]]
 #                     [DEBUG] [EXCLUDE_FROM_ALL]
 #
-# Warning: LINK_LIBRARIES option is a temporary feature and can be removed any 
-# soon. See #113 for further information.
+# Warning: LINK_LIBRARIES option is a temporary feature and can be removed any
+# time. See #113 for further information.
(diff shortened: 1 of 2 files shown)


psigen:
This PR was included into #135 and is no longer necessary.

PR #135, opened 2017-03-30
Overhaul of YAML usage in chimera::CompiledConfiguration.

### LINKED ISSUE #120: Required items in configuration file?
According to the code (e.g., [this](https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp#L280) and [this](https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp#L485)), the items are seemed to be optional. However, `yaml-cpp` 0.5.2 throws exception for the current implementation of Chimera at [here](https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp#L398) (`node.Type()`).

`yaml-cpp` 0.5.2 [creates a *zombie node*](https://github.com/jbeder/yaml-cpp/blob/release-0.5.2/include/yaml-cpp/node/impl.h#L373) if the key doesn't exist (e.g., [`template_config["prebody"]`](https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp#L284)), and the zombie node throws an exception when [`Node::Type()` is called](https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp#L398). This is the cause of the errors reported in #119.

I suggest that we perform the validity check in advance, and, if it's not valid, we print warnings and just return.

psigen:
Follows PR #134.

This PR rewrites most of the YAML access in `CompiledConfiguration` to fix a number of parsing issues related to missing YAML entries, improve error checking and output, and generalize the text-template engine.

* YAML accessors are all result-checked to eliminate many `YAML::InvalidNode` runtime errors.
* Default bindings are now represented via `chimera::Binding` structs, which are resolved once when the configuration is compiled.
* Templates can now use any hierarchy of parameters under the `template::file` and `template:main` namespaces.  File references in these parameters are recursively resolved and automatically forwarded directly to `mstch` templates so we do not have to hardcode parameter names.
* Our poor single unit test now actually generates cpp output and may actually be close to working now.
* I added Boost.Python as a CI dependency so that we can compile our unit test which generates a sort of python binding by default.

There are two minor breaking changes to the YAML configuration file:
1. To include file-snippets in the `template` section, instead of using a map with a key entry of `source:`, we check for a scalar with a type tag of `!file`.  This disambiguates file loading from hierarchical parameters.
2. The `template::cxx_record` entry has been changed to `template::class` because it is a less unwieldy name.

Closes #120.


src/configuration.cpp
```diff
@@ -161,74 +160,154 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
-        auto ns = chimera::util::resolveNamespace(ci, ns_str);
-        if (ns)
+        // Check that 'namespaces' node in configuration YAML is a map.
+        if (!namespacesNode.IsMap())
         {
-            declarations_[ns] = it.second;
-            namespaces_.insert(ns);
+            std::cerr << "'namespaces' in configuration YAML must be a map."
+                      << std::endl;
+            exit(-2);
```
mkoval: Can we throw an exception instead of `exit`ing here? 😭

src/configuration.cpp
```diff
@@ -161,74 +160,154 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+                          << "'" << ns_str << "'." << std::endl;
+                exit(-2);
+            }
         }
     }
 
-    // Resolve class/struct configuration entries within provided AST.
-    for(const auto &it : configNode["classes"])
+    // Parse 'classes' section of configuration YAML if it exists.
+    const YAML::Node &classesNode = configNode_["classes"];
```
mkoval: This is technically undefined behavior because `configNode` returns a `Node` by value. This issue is prevalent throughout Chimera - and a few of my other projects that use `yaml-cpp` 😅 - so it's best to make that change in a followup pull request.

src/configuration.cpp
```diff
@@ -161,74 +160,154 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+        for(const auto &it : typesNode)
         {
-            std::cerr << "Unable to resolve type: "
-                      << "'" << type_str << "'" << std::endl;
-            exit(-3);
+            std::string type_str = it.first.as<std::string>();
+            auto type = chimera::util::resolveType(ci, type_str);
+            if (type.getTypePtrOrNull())
+            {
+                types_.push_back(std::make_pair(type, it.second));
```
mkoval: Nit: I'd prefer to use `emplace_back` here.

src/configuration.cpp
```diff
@@ -161,74 +160,154 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+                          << "'" << type_str << "'" << std::endl;
+                exit(-2);
+            }
         }
     }
+
+    // Construct the binding that is used for this configuration.
+    // If a configuration is not found, the default values will be taken from the preset binding.
+    // TODO: We can now very easily handle multiple default bindings in the
+    //       future by changing the next line to depend on the configuration :)
```
mkoval: 🎉

src/util.cpp
```diff
@@ -37,6 +37,66 @@ std::string generateUniqueName()
...
+    {
+    case YAML::NodeType::Scalar:
+        return fn ? fn(node) : node.as<std::string>();
+    case YAML::NodeType::Sequence:
+        {
+            ::mstch::array context;
+            for(YAML::const_iterator it = node.begin(); it != node.end(); ++it)
+            {
+                const YAML::Node &value = *it;
+                context.push_back(wrapYAMLNode(value));
```
mkoval: Nit: Use `emplace_back` here.

src/util.cpp
```diff
@@ -37,6 +37,66 @@ std::string generateUniqueName()
...
+    for(YAML::const_iterator it=node.begin(); it!=node.end(); ++it)
+    {
+        const std::string name = it->first.as<std::string>();
+        const YAML::Node &value = it->second;
+
+        // Insert the key if overwriting or if the key does not exist.
+        if (overwrite || map.find(name) == map.end())
+        {
+            map[name] = chimera::util::wrapYAMLNode(value, fn);
+        }
```
mkoval: This searches `map` for `name` twice. Replace this with:
```c++
auto result = map.emplace(std::make_pair(name, ::mstch::node{}));
if (result.second || overwrite)
{
    result.first->second = chimera::util::wrapYAMLNode(value, fn);
}
```

src/configuration.cpp
psigen: Created #136 to track this.

src/configuration.cpp
psigen: Created #137 to track this.

PR #138, opened 2017-03-31
Add note about key length to README

### LINKED ISSUE #118: Update README.md for long key case
[`yaml-cpp` doesn't allow more than 1024 characters for a single line key](https://github.com/jbeder/yaml-cpp/blob/master/src/simplekey.cpp#L111). For longer key, we should use [multi-line key](http://stackoverflow.com/a/36295084). It would be good to mention about this on the readme.

mkoval:
I moved the description from the Wiki to the `README`. This fixes #118.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,15 @@ classes:
     noncopyable: true
 ```
 
+## Troubleshooting ##
+
+### Is there a length limit for the keys in the configuration file of Chimera? ###
+
+Yes. [`yaml-cpp` support more than 1024 characters for a single line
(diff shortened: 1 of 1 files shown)


jslee02:
I believe updating `REAME.md` is not the cause of CI failure. :rage:  (see #140).

mkoval:
I am ignoring the CI failure and merging this, since @psigen identified the underlying issue [that is completely unrelated to this pull request].

jslee02:
Are we okay to remove the wiki then?

mkoval:
Done!

PR #142, opened 2017-04-09
Modified CMake to use newer directives.

### LINKED ISSUE #141: Switch to using `target_include_directories` instead of `include_directories`
I didn't realize it, but this is supported in CMake 2.8.12 already.  We should just advance the required version of CMake and use the `target_include_directories` directive as it is cleaner and handles transitive dependencies correctly.

psigen:
This PR updates the CMakeLists.txt to use newer `target_` directives for includes and compile options, which removes some of the cruft associated with inheriting these properties.

The required version of CMake has been updated to 2.8.12 to make use of these newer features. This should be available on all targeted platforms (and specifically, Ubuntu 14.04).

Closes #141

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,7 @@ include_directories(SYSTEM ${CLANG_INCLUDE_DIRS})
 add_definitions(${CLANG_DEFINITIONS})
 
 set(YAMLCPP_STATIC_LIBRARY ON CACHE BOOL
-    "If true, try to find static yaml-cpp library first instead of dynamic.")
+  "If true, try to find static yaml-cpp library first instead of dynamic.")
 find_package(YamlCpp REQUIRED)
 include_directories(SYSTEM ${YAMLCPP_INCLUDE_DIRS})
 add_definitions(${YAMLCPP_DEFINITIONS})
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -44,14 +44,11 @@ endif()
 
 if (CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANGCXX)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra -Wall -Wstrict-aliasing=2")
```
psigen: Should these flags also get moved to the `target_compile_options`?

jslee02: We may want to set the warning flags over all the local targets of Aikido. So `add_compile_options()` would be enough I think.

CMakeLists.txt
```diff
@@ -82,6 +74,15 @@ add_executable("${PROJECT_NAME}"
   src/mstch.cpp
 )
 
+target_compile_options("${PROJECT_NAME}"
+  PUBLIC "--std=c++11"
```
jslee02: I've seen `-std=c++11` (note the single dash) is used to enable C++11 rather than double dashes.

CMakeLists.txt
```diff
@@ -82,6 +74,15 @@ add_executable("${PROJECT_NAME}"
...
+  PRIVATE "-Wno-unused-parameter"
```
jslee02: Unless this is target specific we could move this to `add_compile_options()`.

test/CMakeLists.txt
```diff
@@ -6,21 +6,23 @@ include(chimeraFunctions)
...
 add_chimera_binding(
   EXCLUDE_FROM_ALL
   TARGET test_binding
   SOURCES dart-example.cpp
   NAMESPACES dart
 )
+
+target_include_directories(test_binding SYSTEM
+  PRIVATE ${Boost_INCLUDE_DIRS}
+  PRIVATE ${PYTHON_INCLUDE_DIRS}
```
jslee02: Nit: We don't need to specify `PRIVATE` for each item. This would work:
```cmake
target_include_directories(test_binding SYSTEM
  PRIVATE 
    ${Boost_INCLUDE_DIRS}
    ${PYTHON_INCLUDE_DIRS}
)
```

test/CMakeLists.txt
```diff
@@ -6,21 +6,23 @@ include(chimeraFunctions)
...
+)
+
 target_link_libraries(test_binding
   ${Boost_PYTHON_LIBRARIES}
```
jslee02: Nit: We may want to specify the scope using `PRIVATE`?

jslee02: I didn't test for Boost.Python, but it seems the pattern of boost component variable is `Boost_COMPONENTNAME_LIBRARY`. So this might need to be changed to `Boost_PYTHON_LIBRARY`. Please ignore if I'm wrong, but I confirmed this for [Boost.FileSystem](https://github.com/personalrobotics/aikido/pull/154/files#diff-768ce6c21087a82c0c1a47637a8d8a18R56).

psigen: I think you are right according to: https://cmake.org/cmake/help/v3.0/module/FindBoost.html
```
Boost_<C>_LIBRARY      - Libraries to link for component <C> (may include
                         target_link_libraries debug/optimized keywords)
```

psigen:
Thanks @jslee02!  Good suggestions.  I think I addressed all your comments.

Issue #137, opened 2017-03-31
Remove `YAML::Node&` usage.


psigen:
@mkoval noted in #135 that `YAML::Node&` usage could be unsafe in certain cases, due to the nodes potentially going out-of-scope.  I saw some clang warnings to this effect as well.

We should check our the reference passing and switch to `YAML::Node` in cases where these references could have scope issues.

PR #143, opened 2017-04-09
Support for multiple built-in language bindings.

psigen:
This PR overhauls the binding class stubs added in #135 into a full template generation subsystem to package multiple language bindings.  At the moment, these languages can be selected by defining a `binding: [BINDING_NAME]` key in the configuration YAML, but we can rearrange this as needed later.

Each binding is defined as a set of template files in a subdirectory of `bindings/`.  The `CMakeLists.txt` in this directory iterates through a list of these template files, assembles them into `-impl.h` headers, and then includes those headers into a `binding.cpp` file which is used to build a static library of built-in bindings which is accessible via a global map in `binding.h`.  These `-impl.h` files are then loaded in the constructor of the singleton `Configuration`, and queried in the constructor of `CompiledConfiguration`.

This was the cleanest first pass I could come up that allowed the injection of autogenerated binding definitions without modifying much of the chimera core code.  But I could use some help with some of the rough edges worth ironing out, e.g.:

1. Should `bindings` be a root folder, or should it go inside `src`?  It almost exclusively consists of template files and not source code, but it is still used to generate a `binding.cpp` source file.

2. Is the use of an `extern std::map` in `binding.h` reasonable here, or should there be some sort of dedicated singleton class for querying definitions?  This mitigates the requirement of calling an initialization function in `Configuration`.

3. Do we need to handle weird naming of language bindings?  Currently, I assume each binding's name is also its corresponding subdirectory in `bindings/`.  However, this is probably not the best idea if we want to have nicely formatted text names e.g. `Boost.Python` or `LuaBind++` in the future.  If that's the case, we should probably switch to a tuple list of `config_name / directory_name` in `bindings/CMakeLists.txt`.

4. **Sadly, in the course of this PR, I updated to our detailed DART boost::python templates that breaks our unit test even more by generating a _lot_ of complicated boost::python code that really _really_ doesn't compile instead of the placeholder empty `cpp` classes it previously succeeded with. Not sure how to handle that... :sweat: :cry:**


src/configuration.cpp
```diff
@@ -161,152 +163,172 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
     // Get a reference to the configuration YAML root node.
-    // If it is not available, then skip the remaining parsing.
-    if (!configNode_)
-        return;
-
-    // Parse 'namespaces' section of configuration YAML if it exists.
-    const YAML::Node &namespacesNode = configNode_["namespaces"];
-    if (namespacesNode)
+    // If it is not available, then skip configuration node parsing.
+    if (configNode_)
```
psigen: I didn't actually change any logic here, I just moved the following lines inside the `if` block (up to `L289`), then added the block from `L289` to `L303` where we also check for a `binding: FOO` in the YAML.

psigen:
Ping @jslee02 @mkoval if you get a chance.

bindings/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+# Iterate over each available language binding and generate a C++ source file.
+foreach(binding_NAME IN LISTS binding_LIST)
+    set(binding_IMPL_OUTPUT "${binding_NAME}_binding-impl.h")
+    set(binding_IMPL_PATH "${CMAKE_CURRENT_SOURCE_DIR}/${binding_NAME}")
+
+    # Generate a binding from file template and add to implementation list,
+    # then append the implementation to variables that are used to create a
+    # single binding library.
+    add_custom_command(OUTPUT "${binding_IMPL_OUTPUT}"
+      COMMAND ${CMAKE_COMMAND}
```
mkoval: Nit: Quote `${CMAKE_COMMAND}`.

bindings/binding.cpp.tmpl
```diff
@@ -0,0 +1,16 @@
...
+std::string chimera::binding::DEFAULT_NAME = "boost_python";
+
+// Create the built-in definitions map.
+std::map<std::string, chimera::binding::Definition> chimera::binding::DEFINITIONS;
+
+// Define loader function for all bindings.
+void chimera::binding::initializeBuiltinBindings()
+{
+    @BINDING_REGISTRATIONS_LIST@
+}
```
mkoval: We can eliminate this function by generating an initialization list for `chimera::binding::DEFINITIONS`. I believe that simply requires changing `binding.cpp.tmpl` to contain:
```c++
std::map<std::string, chimera::binding::Definition> chimera::binding::DEFINITIONS {
@BINDING_REGISTRATIONS_LIST@
};
```
And modifying `bindings/CMakeLists.txt` to contain:
```c++
list(APPEND BINDING_REGISTRATIONS_LIST
  { \"${binding_NAME}\", ${binding_NAME}_DEFINITION },
```
You might need to wrap the string in an `std::string`, no warranty, etc.

bindings/binding.cpp.tmpl
```diff
@@ -0,0 +1,16 @@
+// This file was autogenerated by the Chimera build.
```
mkoval: This entire file explicitly prefixes definitions with `chimera::binding::`. This is rather verbose - could we simply put everything inside `namespace` blocks instead?

src/configuration.cpp
```diff
@@ -161,152 +163,172 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
 
-        if (const YAML::Node &moduleTemplateNode = templateNode_["module"])
-            templateBinding_.module_cpp = Lookup(moduleTemplateNode);
+        if (const YAML::Node &moduleTemplateNode = bindingNode_["module"])
+            bindingDefinition_.module_cpp = Lookup(moduleTemplateNode);
 
-        if (const YAML::Node &variableTemplateNode = templateNode_["variable"])
-            templateBinding_.variable_cpp = Lookup(variableTemplateNode);
+        if (const YAML::Node &variableTemplateNode = bindingNode_["variable"])
+            bindingDefinition_.variable_cpp = Lookup(variableTemplateNode);
```
mkoval: I'm pretty sure all of these `YAML::Node`s should be stored by value, not `const &`.

bindings/CMakeLists.txt
psigen: I don't know if we actually want to do this or not, I'm worried someone might pass a `${CMAKE_COMMAND}` which is an executable with flags and should _not_ be treated as one argument.  E.g. `/usr/bin/cmake --my-fun-flag`

src/configuration.cpp
psigen: Agreed, but let's address all of these in a PR to fix #137.

bindings/binding.cpp.tmpl
psigen: I intentionally did this because I was a little worried that the auto-generated variables might somehow contain something that could be scoped incorrectly inside the `chimera::binding` namespace.

Probably unlikely, but it doesn't really get much shorted if you add the `namespace {` blocks in here anyway.  In fact, I'm not sure it saved _any_ space when I tried it.

bindings/binding.cpp.tmpl
psigen: Works great :+1:

psigen:
Ok, I did another pass on this.  Think we can go ahead and merge and address the remaining stuff in followups @mkoval?

I think the highest priority is a PR to address #145 right now.

PR #150, opened 2017-05-19
Added forwarding of source files to templates.

psigen:
This PR adds the ability to forward source file paths to binding templates under the "sources" key.  This functionality is implemented as the following:

1. By default, source paths are collected at startup and stored in the `Configuration`.
2. At render time, the source paths in the `Configuration` are exposed as the template variable "sources".
3. The `--no-default-sources` flag disables this enumeration, making the "sources" variable default to the empty list.
4. If a "sources" key is defined in the YAML configuration for the template, it will override the default autogenerated "sources" key.
5. Both "sources" and "includes" should generally be used by binding templates in `#include` statements.

Resolves #143


src/chimera.cpp
```diff
@@ -51,11 +51,16 @@ static cl::opt<bool> UseCMode(
     "use-c", cl::cat(ChimeraCategory),
     cl::desc("Parse input files as C instead of C++"));
 
-// Option for switching from C++ to C source.
+// Option for preventing auto-extraction of comments from source files.
```
psigen: This was an originally incorrect comment that I just happened to fix.

bindings/boost_python/class.cpp.tmpl
```diff
@@ -48,7 +51,7 @@ void {{class.mangled_name}}()
 {{#class.constructors}}{{!
 }}{{#overloads}}{{!
     }}.def("__init__", ::boost::python::make_constructor({{!
-        }}[]({{#params}}{{type}} {{name}}{{^last}}, {{/last}}{{/params}}){{!
+        }}+[]({{#params}}{{type}} {{name}}{{^last}}, {{/last}}{{/params}}){{!
```
psigen: Adding this '+' operator triggers a conversion of the lambda to a plain function pointer, which resolves compilation issues in boost::python not being able to deduce the type of the lambda function.

See:
http://stackoverflow.com/questions/16845547/using-c11-lambda-as-accessor-function-in-boostpythons-add-property-get-sig
http://stackoverflow.com/questions/18889028/a-positive-lambda-what-sorcery-is-this

PR #152, opened 2017-05-22
Add Codecov support

jslee02:
[Sample report](https://codecov.io/gh/personalrobotics/chimera/branch/codecov).

Depends on #150 and #154


psigen:
@js:  Is this a useful mechanism here?  We are running a generator tool here so even 100% code coverage doesn't indicate whether _any_ boost::python bindings are working, for example.

jslee02:
That's true, but I believe code coverage checking is a necessary condition.

psigen:
That's true, and I suppose we already had Coveralls flags so this isn't a big change.

.travis.yml
```diff
@@ -38,9 +39,10 @@ install:
 script:
   - mkdir build
   - cd build
-  - cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" ..
+  - cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" "-DCODECOV=${CODECOV}" ..
   - make -j4
-  - make test
+  - if [ $CODECOV = ON ]; then make chimera_coverage; else make test; fi
+  - if [ $CODECOV = ON ]; then bash <(curl -s https://codecov.io/bash) || echo "Codecov did not collect coverage reports"; fi
```
psigen: Nit: add `.` at end of echo message.

Issue #148, opened 2017-05-18
Incorrect binding for lambda function parameter


jslee02:
I observed the following binding generated by Chimera:
```cpp
void _ZZN4dart6common7Factory15registerCreatorERKT_EUlDpOT2_E_()
{
::boost::python::object parent_object(::boost::python::scope());
::boost::python::scope parent_scope(parent_object);

::boost::python::class_<(lambda), ::boost::noncopyable >("", boost::python::no_init)
;
}
```

Note that `(lambda)` is generated.

This binding seems to be generated from the following C++ code:
```cpp
template <typename KeyT,
          typename BaseT,
          typename HeldT,
          typename... Args>
template <typename Derived>
void Factory<KeyT, BaseT, HeldT, Args...>::registerCreator(
    const KeyT& key)
{
  return registerCreator(
      key,
      [](Args&&... args) -> HeldT
      {
        return detail::DefaultCreator<Derived, HeldT, Args...>::run(
            std::forward<Args>(args)...);
      }
  );
}
```

jslee02:
A simpler example reproducing the same error is:

**source**:
```cpp
inline std::function<void()> foo_NOT_BINDED()
{
    return [] {};
}
```

**binding**:
```cpp
void _ZZN12chimera_test14foo_NOT_BINDEDEvEUlvE_(::pybind11::module& m)
{
    ::pybind11::class_<(lambda) >(m, "");
}
```

It seems Chimera considers a lambda function that is exposed in the header in a inline function as a class, and the class match fails to extract the class properties (e.g., {{class.type}} is `(lambda)`, and {{class.name}} is empty), which is not surprising as lambda function doesn't have those properties.

jslee02:
By #148, Chimera ignores class mstch if the type is `(lambda)`. We still need to fix this by correctly detecting lambda functions.

PR #158, opened 2017-08-05
[Discussion] Incorrect lambda binding generation

jslee02:
This PR adds a minimal test that reproduces the issue of lambda binding (see #148).


test/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_chimera_binding(
 
 set_target_properties(test_binding PROPERTIES PREFIX "")
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
```
prasbg: Just out of curiosity, why is this flag necessary if the target compile options are set below?  Is there some other target that is not getting the options?

psigen: Just out of curiosity, why is this flag necessary if the target compile options are set below? Is there some other target that is not getting the options?

jslee02: I haven't dig further, but it's not compiled without it complaining `std::function` is not defined even with `<functional>` included.

jslee02:
Closing as this is resolved by https://github.com/personalrobotics/chimera/pull/295

PR #159, opened 2017-10-02
Fixes add_chimera_binding() to inherit target properties.

psigen:
This commit fixes the issue originally raised in #113 where build options for the chimera binding target were not propagated to the placeholder target used in binding generation.

The fix makes use of generator expressions to directly insert the properties for the target into the properties of the placeholder. This works for compile options, include directories, and link libraries, but I don't know if it works quite correctly with transitive dependencies directly on other CMake targets.

I also added a slightly clearer comment to the `empty.cpp` file generated as part of the chimera CMake process, as it is otherwise not clear that it is a placeholder file rather than an incomplete part of a broken build.


jslee02:
I think is a cleaner way than #113. Let me test this my self with `dartpy` to see if it works as expected. If you already confirmed that this works then I'm okay to merge this.

PR #163, opened 2017-10-02
Fixed a typo preventing file source loading.

psigen:
This commit fixes a typo in tag specification that was preventing correct parsing of the `!file` tag type to allow file paths to be specified as values in certain values in the configuration YAML.


psigen:
Ok, I merged this PR because it was 100% wrong before, so we might as well at least run with this fix.

PR #169, opened 2018-05-11
Update test/CMakeLists.txt

jslee02:
**Summary**
- allow to set python version
- find correct boost-python component
- add cmake function to easily add binding tests


psigen:
Thanks @jslee02 , this is some nice code.

The reason I did not do this before is that I was anticipating needing to test `chimera` against multiple python versions and binding libraries, so I didn't want to make the test macro itself contain binding-specific boilerplate.  But it is a lot of boilerplate to be writing over and over.

Maybe what would make sense is to create a `cmake` file that sets up the macros for `Boost_Python` specifically, and then include that and use its exported functions when defining tests.  We might use the same test code under multiple Python versions or with both Boost_Python and PyBind11 in the future, so I could see us creating a single test cpp file and then running:

```
chimera_add_test_boost_python(... SOURCES test.cpp ...)
chimera_add_test_pybind11(... SOURCES test.cpp ...)
```

Do you want to try to do that, or should I try refactor this PR to do that?

jslee02:
@psigen All sounds good to me! Having macros for different bindings would be great. Current macro is specific to the detected python version. I couldn't find a clean way of using different versions of one library in a single CMake project without cleaning cache, which is rather seemed gymnastic of CMake scripts. :sweat: 

Please feel free to refactor PR, or creating new PR would be also fine. :smile:

jslee02:
@psigen I updated this PR based on your comment. I'm merging this as it is because this is blocking other changes. Please let me know if further change is required. Will work on that in separate PRs. 😄

PR #174, opened 2018-05-21
Initial work for pybind11 support

jslee02:
This is very early stage work for supporting pybind11. Currently, only simple function binding is working (`01_function` test).

Edit: Oh, the actual essential initial work was #143. :sweat_smile:


psigen:
Wow, this is awesome!

PR #179, opened 2018-05-27
Pybind11 cleanup

psigen:
This is just a minor refactor and cleanup that simplifies the pybind11 infrastructure created in #174 .

Major changes:
* Refactored test function definitions into a separate `cmake` directory (which helps keep them separate from build configuration)
* Split functions for each binding type into separate cmake files.
* Created new meta-function that creates test for each binding.
* Added ability to specify binding definition at command line.
* Added ability to specify binding definition in CMake `add_chimera_binding()`


test/01_function/CMakeLists.txt
```diff
@@ -1,14 +1,12 @@
-include(chimeraBindingTest)
-
-chimera_add_binding_test_boost_python(function_boost_python
+chimera_add_test_boost_python(function_boost_python
```
jslee02: I used `binding_test` instead of just `test` to distinguish binding tests and possible tests for chimera code itself. I understand it wouldn't be so easy to cover all the chimera code but I think we could have some tests. Let me know what do you think.

src/configuration.cpp
```diff
@@ -149,7 +172,7 @@ chimera::CompiledConfiguration::CompiledConfiguration(
 , ci_(ci)
 {
     // Start out by setting the binding name to the default.
```
jslee02: Nit: This comment needs to be updated to something like "Start out by setting the binding name to an empty string.".

Issue #168, opened 2018-05-11
Unable to generate bindings for non-member functions not defined in header


jslee02:
Here is a simple example:
```cmake
// CMakeLists.txt
add_chimera_binding(
  EXCLUDE_FROM_ALL
  TARGET test_binding
  SOURCES dart-example.h # adding dart-example.cpp complaints multiple definitions in the binding files.
  NAMESPACES dart
)
```
```cpp
// dart-example.h
void foo(); // no binding!
inline void bar() {} // okay

// dart-example.cpp
void foo() {}
```

psigen:
Thanks! This is much easier to test with!!

jslee02:
Further observation: If I pass `dart-example.cpp` to SOURCE then the binding of `foo()` is generated, but the compiler complains about multiple definitions as:
```shell
$ cmake ../../..; make -j12 test_binding
-- Clang libs: <...>
-- Found Clang: /usr/lib/llvm-4.0/include
-- Boost version: 1.58.0
-- Boost version: 1.58.0
-- Found the following Boost libraries:
--   python-py35
--   thread
--   chrono
--   system
--   date_time
--   atomic
-- CHIMERA_TEST_Boost_PYTHON_LIBRARIES: /usr/lib/x86_64-linux-gnu/libboost_python-py35.so
-- PYTHON_LIBRARIES: /usr/lib/x86_64-linux-gnu/libpython3.5m.so
-- Configuring done
-- Generating done
-- Build files have been written to: <...>/chimera/master/build/GCC/Release
[ 22%] Built target mstch_EXTERNAL
[ 31%] Built target chimera_bindings
[ 37%] Built target cling_utils
[ 60%] Built target chimera
[ 62%] Built target test_binding_SOURCES
[ 85%] Built target test_binding_REBUILD
Scanning dependencies of target test_binding
[ 88%] Building CXX object test/CMakeFiles/test_binding.dir/test_binding/empty.cpp.o
[ 91%] Building CXX object test/CMakeFiles/test_binding.dir/test_binding/_ZN4dart6common5FrameE.cpp.o
[ 94%] Building CXX object test/CMakeFiles/test_binding.dir/test_binding/_ZN4dart6common3fooEi.cpp.o
[ 97%] Building CXX object test/CMakeFiles/test_binding.dir/test_binding/test_binding.cpp.o
[100%] Linking CXX shared module test_binding.so
[...]

PR #183, opened 2018-05-27
Added python tests in unit testing.

psigen:
This PR adds `add_test()` directives that run python unit tests against the generated code.  These were previously available, but not run directly through CTest.

I also modified the code in `01_function` to exacerbate the bug in #168, so we will need to fix that bug before unit tests will pass again.


psigen:
Ok, this may have fixed #168, I am not sure.  The unit tests are now passing.

jslee02:
I tested this with `dartpy`'s `dart-6.5` branch. It seems this change generates a function with not fully qualified argument types. For example, there is a generated binding, which wasn't generated before this change, that includes not fully qualified type:
```cpp
void _ZN4dart6commonL26getVectorObjectIfAvailableIPNS_8dynamics4NodeEEET_mRKSt6vectorIS5_SaIS5_EE()
{
::boost::python::object parent_object(::boost::python::scope().attr("common"));

::boost::python::def("getVectorObjectIfAvailable", [](std::size_t index, const std::vector<Node *> & vec) -> dart::dynamics::Node * { return dart::common::getVectorObjectIfAvailable<dart::dynamics::Node *>(index, vec); }, ::boost::python::return_value_policy<::boost::python::return_by_smart_ptr<dart::dynamics::NodePtr> >(), (::boost::python::arg("index"), ::boost::python::arg("vec")));

}
```
Note that `const std::vector<Node *>` should be `const std::vector<dart::dynamics::Node *>`.

psigen:
Hmm, do you think there is any way to make a unit test for this case?

psigen:
> Do we need to add all the source files to EXTRA_SOURCES option to get bindings of non-member functions?

No, we do not.  The purpose of the `EXTRA_SOURCES` option is to allow us to include `cpp` files that _should_ be compiled into the binding, but should _not_ be used to generate the binding.  This can be useful if we have various helper function implementations or similar things.

jslee02:
> Hmm, do you think there is any way to make a unit test for this case?

Let's handle this case in a separate PR. Let me create an issue for this.

PR #192, opened 2018-09-14
Ignore pybind11 binding tests when not built

### LINKED ISSUE #188: Python tests for pybind11 can fail when pybind11 isn't installed
By #183, we tests the generated python modules for both of Boost.Python and pybind11, but pybind11 test can fail when pybind11 isn't installed. This is because the pybind11 bindings are not generated in that case. We need to either make pybind11 required dependency for tests or modify the pybind11 tests optional.

jslee02:
resolves #188

### DIFF
--- a/test/01_function/function.py
+++ b/test/01_function/function.py
@@ -1,7 +1,11 @@
 import unittest
 
-from function_pybind11.nested_function import add as add_py11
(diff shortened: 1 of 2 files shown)


jslee02:
Let me merge this PR because this will fixes the binary packages for PRL PPA. Please feel free to give feedback even after the merge.

PR #193, opened 2018-09-14
Support class inheritance and static function for pybind11

jslee02:


CMakeLists.txt
```diff
@@ -81,6 +92,7 @@ add_subdirectory(external)
 
 # Build the main chimera executable.
 add_executable("${PROJECT_NAME}"
+  ${chimera_headers}  # not required, but helpful for some IDEs
```
psigen: If we are going to do this, let's clean it up and make the variables
 `"${PROJECT_NAME}_HEADERS"` and `"${PROJECT_NAME}_SOURCES"` variables and use those in the `add_executable()` block instead of handling the two differently.

jslee02:
@psigen Is this PR good to go?

PR #195, opened 2018-09-15
Support class static fields for pybind11

jslee02:
depends on #193


jslee02:
This change newly introduce following error when building dartpy for DART 6.6:
```
chimera: /usr/lib/llvm-5.0/include/clang/AST/DeclCXX.h:604: clang::CXXRecordDecl::DefinitionData& clang::CXXRecordDecl::data() const: Assertion `DD && "queried property of class with no definition"' failed.
```
Not sure if this is a llvm 5 specific error or not.

jslee02:
The error disappears by disabling static fields generation for DART. I think this error is just unveiled as we generate bindings for static fields of DART, but it shouldn't block merging this PR.

Issue #194, opened 2018-09-15
Add make_setter if static property is assignable


jslee02:
See: https://github.com/personalrobotics/chimera/blob/master/bindings/boost_python/class.cpp.tmpl#L98

PR #196, opened 2018-09-15
Add smart pointer tests

jslee02:
depends on #194


CMakeLists.txt
```diff
@@ -66,6 +66,7 @@ endif()
 # Globally inherit the top-level include directory because it is difficult to
 # locate from CMakeLists.txt in source subdirectories.
 include_directories("${PROJECT_SOURCE_DIR}/include")
+file(GLOB_RECURSE chimera_headers "${PROJECT_SOURCE_DIR}/include/*.h")
```
prasbg: I would prefer not to compute these headers using a GLOB_RECURSE, because it sometimes leads to invalid caching if the files are changed.

CMakeLists.txt
```diff
@@ -81,6 +82,7 @@ add_subdirectory(external)
 
 # Build the main chimera executable.
 add_executable("${PROJECT_NAME}"
+  ${chimera_headers}
```
prasbg: Why do we need to do this here?  Is there a platform that requires the headers as sources?

test/02_class/CMakeLists.txt
```diff
@@ -1,21 +1,14 @@
...
-  CONFIGURATION ${CMAKE_CURRENT_SOURCE_DIR}/test_class_boost_python.yaml
-  DEBUG
-  COPY_MODULE
-)
-
-chimera_add_binding_test_pybind11(test_class_pybind11
-  SOURCES test_class.cpp
-  NAMESPACES test_class
-  CONFIGURATION ${CMAKE_CURRENT_SOURCE_DIR}/test_class_pybind11.yaml
+chimera_add_binding_tests(class
```
prasbg: 👍

CMakeLists.txt
jslee02: I wished you missed this change but failed! Maybe I'm too lazy maintaining the source list manually. :/

Let me revert this change.

CMakeLists.txt
jslee02: Not required but useful when using Qt Creator. Open to remove this.

CMakeLists.txt
psigen: https://cmake.org/cmake/help/v3.3/command/file.html?highlight=glob#file
>Note:  We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

jslee02: Fixed in #193 since this PR include the changes of #193.

jslee02: Yeah, I know that CMake also officially doesn't recommend that. I simply don't prefer that way. But I'm okay with not to use it though.

PR #199, opened 2018-09-15
Remove unused private member

jslee02:


jslee02:
@psigen The private member `class_decl_` isn't used in `Parameter` class.  Can we remove this and change the constructor of `Parameter` not to take `class_decl` as well?

psigen:
I was a bit torn on this.

The reason that I was leaving in this argument was that I wanted to maintain parity with the _other_ type classes which build up their parameters in the same way.  For example, it seemed like this should have at _least_ the same arguments as a `Method`, since this is even _more_ specific in scope.  Also I wasn't convinced that we might not need this argument to access some property in the future.

Would it be possible to mark it unused via a macro, similar to the strategy in DART and Aikido?

jslee02:
That make sense. Let's leave it as it is.



> Would it be possible to mark it unused via a macro, similar to the strategy in DART and Aikido?

We used [this tactic](https://stackoverflow.com/a/1486931/3122234), but I don't know if there is a way for a member variable.

psigen:
@jslee Perhaps we can apply the `UNUSED` macro on both the constructor arg and the member variable inside the constructor:
```c++
Parameter::Parameter(const ::chimera::CompiledConfiguration &config,
                     const ParmVarDecl *decl,
                     const FunctionDecl *method_decl,
                     const CXXRecordDecl *class_decl,
                     const std::string default_name)
: ClangWrapper(config, decl)
, method_decl_(method_decl)
, class_decl_(class_decl)  // <--- Let's add this line
, default_name_(default_name)
{
    // Reserved for future use.
    UNUSED(class_decl_);  // <-- Should mark member as "used"?

    register_methods(this, {
        {"name", &Parameter::name},
        {"type", &Parameter::type},
    });
}
```

PR #201, opened 2018-09-16
Support enumeration for pybind11

jslee02:
Caveat: nested enumeration (probably class) doesn't work with current pybind11 templates.

### DIFF
--- a/test/04_enumeration/enumeration.py
+++ b/test/04_enumeration/enumeration.py
@@ -14,14 +14,14 @@ class TestFunction(unittest.TestCase):
     def test_function_py11(self):
         if not has_pybind11:
             return
-        
-        animal = py11.Animal(py11.Cat)
-        self.assertEqual(animal.type, py11.Cat)
+
+        animal = py11.Animal(py11.Animal.Cat)
+        self.assertEqual(animal.type, py11.Animal.Cat)
 
 
     def test_function_bp(self):
-        animal = boost.Animal(boost.Type.Cat)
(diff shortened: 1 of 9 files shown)


test/04_enumeration/enumeration.py
```diff
@@ -0,0 +1,28 @@
+import unittest
+
+try:
+    import enumeration_pybind11.nested_namespace as py11
+    has_pybind11 = True
+except:
+    has_pybind11 = False
+
+import enumeration_boost_python.nested_namespace as boost
```
psigen: I think I missed this in the other PR.  Where do the `enumeration_xx_yy` namespaces come from here?

psigen:
In general, I am fine with adding more functionality to these pybind11 templates even if they are not complete, as long as we do not introduce regressions.  Each feature makes chimera more useful!

test/04_enumeration/enumeration.py
jslee02: That's the name of the binding, which is specified at [here](https://github.com/personalrobotics/chimera/blob/py11_enum/test/04_enumeration/CMakeLists.txt#L3) with suffixes depending on the binding library (`_boost_python` or `_pybind`).

jslee02:
I started to switch the `dartpy` bindings to pybind and am adding necessary features to here whenever I find something missing. I finished `dart::common` and started `dart::math`. Once `dartpy` completely migrate to pybind11 with full API of dart, I'm pretty sure, pybind11 support of chimera will be reasonably sufficient.

PR #203, opened 2018-09-16
Support variable for pybind11

jslee02:


test/05_variable/variable.py
```diff
@@ -0,0 +1,25 @@
+import unittest
+
+try:
+    import variable_pybind11.nested_namespace as py11
```
psigen: I may have already asked this, but how do we get these top-level namespaces per each binding type?

PR #208, opened 2018-09-18
Add dependency options to test macros

jslee02:


jslee02:
Glad to get two approvals

PR #209, opened 2018-09-18
Mark override overridden virtual functions

jslee02:
(I accidentally committed for this change to master branch so reverted it :sweat: )


jslee02:
I'm using [Clang Code Model integrated with Qt Creator](http://blog.qt.io/blog/2018/06/05/qt-creators-clang-code-model/). AFAIK, we could use a linter to detect them, but fixing them is manual, isn't it?

psigen:
Yup, I was just going to ask for the settings.

PR #210, opened 2018-09-19
Fix for segfault with void* parameter with LLVM >= 3.9

jslee02:
This PR is for bug reporting. chimera segfaults with `void *` parameter.

If the parameter type is changed to something else (e.g., `void` or `int`), the error goes away.


jslee02:
Interesting, it seems the error doesn't occur with LLVM 3.6. I tested with LLVM 5.0 + GCC 7.3.0.

jslee02:
Investigating further, I found that chimera + LLVM 3.6 prints warning for this case as:
```
Warning: Skipped function '<function_name> because the parameter at index 0 has an incomplete type 'void *'.
```
while other LLVM (> 4.0) versions just segfault.

psigen:
Does this mean that it would be impossible to generate bindings to a function that takes a `void *` argument or returns `void *`?

jslee02:
It was skipped anyway even without this change so I keep the behavior as it was, but I'm not sure if that's the correct behavior. If we don't skip the case, the binding is generated anyway, and the `void *` type is binded as Python `Capsule` type in pybind11 (let me double check Boost.Python). I don't know how to use Capsule type, but it seems that's the correct conversion according to [this doc](https://docs.python.org/3.4/extending/extending.html#providing-a-c-api-for-an-extension-module). That being said, we shouldn't skip `void *` types.

By the way, this check seems to be used only for parameter types because I could still generate [this void return function](https://github.com/personalrobotics/chimera/pull/210/files#diff-6883b728ab51ac4bf87c7729a29c92f0R13).

jslee02:
Alright, I changed the logic, and `pybind` works fine with void pointer (see [this](https://github.com/personalrobotics/chimera/pull/210/files#diff-77d72da989961f73994685bbc4c1ec15R21)).

prasbg:
So is the current status that `pybind`does work, but `boost.python` does not work, or that both work now?

jslee02:
void parameter: both work
void pointer parameter: only pybind11 works

src/util.cpp
```diff
@@ -502,6 +502,14 @@ bool chimera::util::containsIncompleteType(Sema &sema, QualType qual_type)
         const ReferenceType *reference_type = cast<ReferenceType>(type);
         return containsIncompleteType(sema, reference_type->getPointeeType());
     }
+    else if (type->isVoidPointerType())
+    {
+      return false;
```
psigen: nit: fix indentation here to match rest of file.

src/util.cpp
```diff
@@ -502,6 +502,14 @@ bool chimera::util::containsIncompleteType(Sema &sema, QualType qual_type)
...
+    }
+    else if (type->isVoidType())
+    {
+      return false;
```
psigen: nit: fix indentation here to match rest of file.

PR #211, opened 2018-09-19
Add QUIET option to cmake functions

jslee02:


cmake/chimeraFunctions.cmake
```diff
@@ -55,31 +55,33 @@ function(add_chimera_binding)
     # If the debug flag is specified, dump all the chimera settings.
     if(binding_DEBUG)
```
psigen: Wait a second.. is the only difference between the `DEBUG` and `QUIET` options here one print statement `message(STATUS "Chimera binding: ${binding_TARGET}")`?

If so, we should eliminate one of the options or change what it means.  Because instead of _adding_ the `QUIET` flag, I would have normally _removed_ the `DEBUG` flag, because we already _only_ print when that flag is specified.

@jslee02 : what do you think the `DEBUG` flag should be for?  I was using it to indicate that we should print a lot of debugging information.

jslee02: Oh, I didn't pay attention to `DEBUG` option, and they look redundant.

I think the current behavior with `DEBUG` option is good, but it would be still useful to see the list of tests (without the full debug information). Maybe we could use CMake property that just has the test list, which is collected by the test macro, and we print the list separately?

prasbg: I think we should just add a separate print to add_binding_tests to print the test names, that's not really the intent of the `DEBUG` flag for `add_chimera_binding()` anyway.

jslee02: Agreed. Let me make that change in another PR.

jslee02: How about we add an option to control the verbosity of tests? For example, if we set `CHIMERA_VERBOSE=ON`, then we turn on DEBUG and don't print the test list, and if `CHIMERA_VERBOSE=OFF` then we turn off `DEBUG` and print the test list.

PR #214, opened 2018-09-20
Add "is_return_type_void" method to Function

jslee02:
chimera always generates lambda function that has trailing return type and return keyword even for void type. Some compiler complains about this.

This PR adds a method to determine whether the return is void and updates the templates accordingly to not have lambda function returning void.

C++ code:
```cpp
void void_return() {}
```

Binding (before):
```cpp
sm.def("void_return", +[]() -> void { return void_return(); });
```

Binding (after):
```cpp
sm.def("void_return", +[]() { void_return(); });
```


psigen:
I am generally ok with this, but I would like to know more about why the compiler is complaining here.  Do you have the specific error message?

We may want to change the terminology here to `isVoidFunction()` or potentially just `isVoid()` if the compiler error message is _"error: void function 'foobar' should not return a value"_ to help make it clearer what this means.

The reason to use the shortened `isVoid()` is because we are already referring to a Function object, ["void function" is a fairly common term](https://www.google.com/search?q=%22void+function%22), and this means we can create a corresponding `isVoid()` accessor on Method, which also has this attribute and might need to expose it for similar reasons.

psigen:
If that is too terse, we could also mirror Clang's existing API for type with:
[isVoidType](https://clang.llvm.org/doxygen/classclang_1_1Type.html#a08a8d6c5fb67e56ace4f2ce305701781) and [isVoidPointerType](https://clang.llvm.org/doxygen/classclang_1_1Type.html#a27a66e13de0f33a48523e25961ded5e7) and do *isVoidReturnType*.  But we should still make sure it works similarly for both Function and Method cases.

jslee02:
I'm okay with the suggested names. I would prefer `isVoid()`, but others sound good too.

> Do you have the specific error message?

I had, but for some reason, I cannot reproduce the error. Let me test further.

Issue #216, opened 2018-09-22
Fails to infer template parameter type in std::vector<T> when T is a pointer type


jslee02:
Here is a sample code:
```cpp
#include <vector>

namespace chimera_test {
namespace common {

template <typename T>
inline void vector_param(T val) { }

template <typename T>
inline void vector_param(const std::vector<T>& vec) { }

class Node
{
public:
  Node() = default;
};

class Skeleton
{
public:
  Skeleton() = default;
  inline void foo()
  { 
    auto vec = std::vector<Node>();
    auto ptrVec = std::vector<Node*>();
    
    common::vector_param(vec);  // ok

    // error: ‘Node’ was not declared in this scope
    // sm.def(
    //     "vector_param",
    //     +[](const std::vector<Node *> & vec) -> void 
    //         { return chimera_test::common::vector_param<chimera_test::common::Node *>(vec); },
    //     ::pybind11::arg("vec"));
    common::vector_param(ptrVec);

    common::vector_param(Node());      // ok
    common::vector_param(new Node());  // ok
  }
};

} // namespace dynamics
} // namespace chimera_test
```

Note that the scope of `Node` in `const std::vector<Node *>&` isn't parsed.

This happens for both of Boost.Python and pybind11 where Boost.Python additionally complains about the held type.

jslee02:
Update:
Investigating further, [this line](https://github.com/personalrobotics/chimera/blob/master/external/cling/src/cling_utils_AST.cpp#L1371) transforms:
```cpp
vector<chimera_test::nested_namespace::Node>
vector<chimera_test::nested_namespace::Node *>
```
to
```cpp
std::vector<chimera_test::nested_namespace::Node>
std::vector<Node *>  // wrong !
```

jslee02:
Resolved by #220

PR #220, opened 2018-10-07
Possible solution for #216

jslee02:
This PR is a possible solution for #216.

~~The underlying reason of #216 was that a pointer type isn't regarded as a tag type (see [here](https://github.com/personalrobotics/chimera/pull/220/files#diff-d46863eecd20f5318713c5a61b126a34R1516)). However, I'm not sure if this is a correct way to fix this.~~ :disappointed:

Update: I've found this fix by trial-and-error without complete understanding.

**Caveat**
~~This solution doesn't cover the case that the template parameter has `const` specifier. See [this](https://github.com/personalrobotics/chimera/pull/220/files#diff-db7bca1250885e1e820db68282d10d90R44) for the detail.~~
Update: This is resolved by https://github.com/personalrobotics/chimera/pull/220/commits/5bbaf6f82c5a11983665815ef4fdf1f927645749.


jslee02:
@psigen Could you take a look at this change when you have a chance?

psigen:
@jslee02 when I was looking through the changes to AST.cpp to make this work, I noticed that the corresponding block of code was actually modified upstream: https://github.com/root-project/cling/blob/master/lib/Utils/AST.cpp#L1653

I am taking a look at what the impact of using the latest `cling/Utils/AST.cpp` is here: #221.
It seems like it may be having a similar effect, but I am not sure.

PR #223, opened 2018-10-14
Use docker for testing against recent Ubuntu distros

jslee02:


.ci/install_macos.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-set -x
+set -e
 # Add e option once we move to using Brewfile for installing packages.
```
psigen: Can we remove this comment since this starts using `-e`?

.travis.yml
```diff
@@ -8,30 +8,93 @@ matrix:
...
+        - COMPILER=CLANG
+        - LLVM_VERSION=5
     - os: osx
       compiler: clang
-      env: BUILD_TYPE=Release CODECOV=OFF COMPILER=CLANG LLVM_VERSION=6
+      env:
+        - BUILD_NAME=MACOS_CLANG_LLVM_6_RELEASE
+        - BUILD_TYPE=Release
+        - COMPILER=CLANG
+        - LLVM_VERSION=6
```
psigen: Wow! So many platforms :heart_eyes:

.travis.yml
```diff
@@ -8,30 +8,93 @@ matrix:
...
+
+before_install:
+  - if [ -n "$DOCKER_FILE" ]; then
+      docker build -t "$DOCKER_FILE" -f ".ci/docker/$DOCKER_FILE" .;
+      docker run -itd -v $TRAVIS_BUILD_DIR:$TRAVIS_BUILD_DIR --env-file .ci/docker/env.list --name chimera-docker "$DOCKER_FILE";
```
psigen: Do you need to explicitly shutdown this container or use the `--rm` flag or will TravisCI just handle the cleanup?

.ci/docker/ubuntu-bionic
```diff
@@ -0,0 +1,18 @@
+FROM ubuntu:bionic
```
psigen: Let's either use the prefix or suffix `Dockerfile` here to indicate this filetype.
`Dockerfile.ubuntu-bionic` is one common convention.

ci/script.sh
```diff
@@ -7,7 +7,7 @@ cd build
 cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" "-DCODECOV=${CODECOV}" ..
```
psigen: There appear to be two versions of this script in `\.ci` and `\ci`?

.travis.yml
```diff
@@ -8,30 +8,93 @@ matrix:
...
+      services: docker
+    - os: linux
+      compiler: gcc
+      env:
+        - BUILD_NAME=BIONIC_GCC_RELEASE
+        - BUILD_TYPE=Release
+        - COMPILER=GCC
+        - LLVM_VERSION=3.9
+        - PYTHON_VERSION=3.6
+        - DOCKER_FILE="ubuntu-bionic"
```
psigen: This is conventionally spelled as one word e.g. `DOCKERFILE`

.ci/install_macos.sh
jslee02: Right, the comment and option should be consistent.

More importantly, I'm still investigating whether this option causes the build failures on macOS (I'm testing it in `docker_v2_test` branch).

jslee02:
* Addressed the comments
* Added Azure Pipeline (alternative CI service)
* Disabled building on macOS because of an unrelated issue.

@psigen Could you do a second pass for this PR?

PR #225, opened 2018-10-14
Ignore unresolvable configuration instead of segfaulting

jslee02:
This is a suggestion rather than fixing a bug.

Currently, chimera raises a segfault when it found an unresolvable declaration in the configuration file. However, sometimes a chimera project could have different configurations depending on the platform, and the project could see a configuration that is not invalid on some platforms.

This can be resolved by the chimera project (e.g., using different configurations per platform), but I think it would be still safe to ignore unresolvable declarations in the configuration file with warnings.

---

This CR depends on https://github.com/personalrobotics/chimera/pull/263

### DIFF
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -78,6 +78,16 @@ public:
     void AddSourcePath(const std::string &sourcePath);
 
     /**
+     * Set whether to treat unresolvable configuration as errors.
+     */
+    void SetStrict(bool val);
+
+    /**
+     * Return whether to treat unresolvable configuration as errors.
+     */
+    bool GetStrict() const;
+
+    /**
      * Process the configuration settings against the current AST.
      */
     std::unique_ptr<CompiledConfiguration> Process(
@@ -114,6 +124,7 @@ protected:
     std::string outputModuleName_;
     std::vector<std::string> inputNamespaceNames_;
     std::vector<std::string> inputSourcePaths_;
+    bool strict{false};
 
     friend class CompiledConfiguration;
 };
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -38,78 +38,73 @@ public:
     Configuration &operator=(const Configuration &) = delete;
 
     /**
-     * Get the chimera configuration singleton for this process.
+     * Gets the chimera configuration singleton for this process.
      */
     static Configuration &GetInstance();
 
     /**
-     * Load the specified file to use as the YAML configuration.
+     * Loads the specified file to use as the YAML configuration.
      */
     void LoadFile(const std::string &filename);
 
     /**
-     * Set the desired binding definition by name.
+     * Sets the desired binding definition by name.
      * If unspecified, the default is "boost_python".
      */
     void SetBindingName(const std::string &name);
 
     /**
-     * Set the desired output path to be prepended to every binding.
+     * Sets the desired output path to be prepended to every binding.
      * If unspecified, the default is the current working directory.
      */
     void SetOutputPath(const std::string &path);
 
     /**
-     * Set the desired output python module for the top-level binding.
+     * Sets the desired output python module for the top-level binding.
      * If unspecified, the default is "chimera_binding".
      */
     void SetOutputModuleName(const std::string &moduleName);
 
     /**
-     * Append an enclosing namespace that should be generated as part
+     * Appends an enclosing namespace that should be generated as part
      * of the top-level binding.
      */
     void AddInputNamespaceName(const std::string &namespaceName);
 
     /**
-     * Append the path to a source file that will be processed as part
+     * Appends the path to a source file that will be processed as part
      * of the binding generation.
      */
     void AddSourcePath(const std::string &sourcePath);
 
     /**
-     * Set whether to treat unresolvable configuration as errors.
+     * Sets whether to treat unresolvable configuration as errors.
      */
     void SetStrict(bool val);
 
     /**
-     * Return whether to treat unresolvable configuration as errors.
-     */
-    bool GetStrict() const;
-
-    /**
-     * Process the configuration settings against the current AST.
+     * Processes the configuration settings against the current AST.
      */
     std::unique_ptr<CompiledConfiguration> Process(
         clang::CompilerInstance *ci) const;
 
     /**
-     * Get the root node of the YAML configuration structure.
+     * Gets the root node of the YAML configuration structure.
      */
     const YAML::Node &GetRoot() const;
 
     /**
-     * Get the filename of the loaded YAML configuration file, if it exists.
+     * Gets the filename of the loaded YAML configuration file, if it exists.
      */
     const std::string &GetConfigFilename() const;
 
     /**
-     * Get the desired output path for bindings.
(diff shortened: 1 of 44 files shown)


psigen:
This is a topic that @mkoval and I debated quite a bit.

At the core of the issue is that if you are not very diligent, this is a very easy way for small typos in your configuration to go undetected.  Chimera will simply ignore the definition and move on with its default behavior.  Early in development this led to us accidentally missing entire modules worth of configuration.

On the other hand, if we are selectively compiling modules, it is also very inconvenient to run a preprocessor stage on the YAML file to dynamically control what modules are present or absent.  This can itself introduce errors by incorrectly generating configuration in the first place.

At the end of the day, this is probably a situation where --although I really hate to do it-- we should have a flag to switch the desired behavior between warnings and errors.  Since we are using clang, I wonder if there is an existing `-W` flag that we can co-opt.

jslee02:
> At the end of the day, this is probably a situation where --although I really hate to do it-- we should have a flag to switch the desired behavior between warnings and errors. Since we are using clang, I wonder if there is an existing `-W` flag that we can co-opt.

Maybe we can specify this policy in the configuration file?

psigen:
Yes, perhaps we can add a `strict` config and matching command line option.

jslee02:
Sounds good to me!

jslee02:
> Yes, perhaps we can add a `strict` config and matching command line option.

Added command line option `-strict`, but not for config yet.

What would be the best location for `strcit` config? I found `arguments` in some example configs something like:
```yaml
arguments:
  - "-extra-arg"
  - "-I/usr/lib/clang/3.6/include"
```
I guess adding `-strict` to the `arguments` tag should work, but I wasn't able to find code does the job.

psigen:
@jslee02 I think you want to implement a block like the following (this is how `bindingName` is resolved from multiple locations): 
https://github.com/personalrobotics/chimera/blob/887dbbc7f30491933f3b004eca136db15677c738/src/configuration.cpp#L407

include/chimera/configuration.h
```diff
@@ -77,6 +77,16 @@ class Configuration
...
 
+    /**
+     * Set whether to treat unresolvable configuration as errors.
+     */
+    void SetStrict(bool val);
+
+    /**
+     * Return whether to treat unresolvable configuration as errors.
+     */
+    bool GetStrict() const;
```
psigen, line 88: Remove the `GetStrict()` from here and put it into the `CompiledConfiguration` instead.  This will give you a chance to resolve it in the `CompiledConfiguration` locally, use it there, then expose it as a getter `GetStrict()` for any other consumers.

jslee02:
> I think you want to implement a block like this in the `CompiledConfiguration` constructor (this is how `bindingName` is resolved from multiple locations):

Sounds good to me. Do you think we should implement for parsing `arguments` tag, which I believe it's not implemented, and parse `-strict` from there? Or we create a new root tag something like following?
```yaml
options:
  strict: True
  # add more options
```

psigen:
I think we should use a new root tag, the arguments are really for clang
options, not chimera ones.

jslee02:
> I think we should use a new root tag, the arguments are really for clang options, not chimera ones.

Done in https://github.com/personalrobotics/chimera/pull/225/commits/2640cea5600bcb025e2c8d03e6c9725b7eb61420

jslee02:
bump :smile:

src/configuration.cpp
```diff
@@ -172,6 +179,48 @@ chimera::CompiledConfiguration::CompiledConfiguration(
   , bindingNode_(configNode_["template"]) // TODO: is this always ok?
   , ci_(ci)
 {
+    // This placeholder will be filled in by the options.strict specified
+    // in the configuration YAML if it exists, or remain unset otherwise.
+    boost::optional<bool> config_options_strict;
+
+    // Parse 'options' section of configuration YAML if it exists.
+    if (configNode_)
```
psigen, line 187: Do we have to do this explicitly?  I thought `YAML::Node` would simply return a node of type `Undefined` on a lookup failure, which would allow you to do optional chaining e.g.:

```
const auto &strictNode = configNode_["options"]["strict"];
if (strictNode_ )
{
    [...]
}
```

src/configuration.cpp
```diff
@@ -172,6 +179,48 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+    // Set the options.strict from one of the following sources in order of
+    // priority: 1) CLI '-strict' setting (True if -strict passed, False
+    // otherwise) 2) YAML configuration setting (True/False) 3) false by default
+    if (parent.strict_)
+    {
+        strict_ = parent.strict_;
+    }
+    else if (config_options_strict)
+    {
+        strict_ = *config_options_strict;
```
psigen, line 217: If you can do optional chaining, then the YAML logic can probably just go in here, since it will be much shorter.

src/configuration.cpp
jslee02: `YAML::Node` throw an exception for lookup from a "zombie" (invalid) node. In this case, `configNode_["options"]` could be a zombie node when "options" tag is missing.

psigen: Ugh.  Can we make a lookup helper that just something like:
```
YAML::Node& safeLookup(const YAML::Node &node, const char *key)
{
    return node[key];
}

template<typename... Args>
YAML::Node& safeLookup(const YAML::Node &node, const char *key, Args... args)
{
    const auto &next = node[key];
    if (!next) return next;
    safeLookup(next, args...);
}

const auto &result = safeLookup(configNode_. "options", "strict");
```

We don't have to do this if it's too difficult, we can approve as is too.

psigen: This would go into e.g. https://github.com/personalrobotics/chimera/blob/887dbbc7f30491933f3b004eca136db15677c738/include/chimera/util.h if we implemented it

jslee02: Done in https://github.com/personalrobotics/chimera/pull/225/commits/b40af517af75a8c537d889310dd49bd1af21da0e.

Modified to return invalid node for both function overloads when `node` itself is invalid. Otherwise, those functions would throw an exception.

src/configuration.cpp
jslee02: We still need to have the YAML logic out of this if-statement because we should check if (1) the tag is specified and (2) the YAML node is a scalar type before this if-statement.

include/chimera/util.h
```diff
@@ -46,6 +46,35 @@ void extendWithYAMLNode(::mstch::map &map, const YAML::Node &node,
...
+
+    auto next = node[key];
+
+    // Return if failed to find a tag of 'key'
+    if (not next)
+        return next;
+
+    // Lookup for the next nested tags
+    return lookupYAMLNode(next, std::forward<Args>(args)...);
+}
```
jslee02, line 76: @psigen If you don't prefer in-line function definitions, then we could consider moving the definition to the bottom of this header (keeping the declaration here) or to a new header (e.g., `util_impl.h`)

psigen: I'm ok with this as is, there isn't much in this file.

Issue #226, opened 2018-10-14
Feature request: Ignore declarations of specific namespaces


jslee02:
It would be great if we can ignore all the declarations of specific namespaces such as detail namespaces instead of specifying all the individual declarations to be ignored.

PR #227, opened 2018-10-19
Allow to skip specific namespaces

jslee02:
This addresses the requested feature in #226.

### DIFF
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -139,7 +139,7 @@ public:
      * is desirable since parent namespaces will naturally be lexically
      * ordered before their children.
      */
-    const std::set<const clang::NamespaceDecl*>& GetNamespaces() const;
+    const std::set<const clang::NamespaceDecl*>& GetNamespacesIncluded() const;
 
     /**
      * Returns list of namespace declarations that should be skipped.
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -141,7 +141,10 @@ public:
      */
     const std::set<const clang::NamespaceDecl*>& GetNamespaces() const;
 
-    const std::set<const clang::NamespaceDecl*>& GetNamespaceBlacklist() const;
+    /**
+     * Returns list of namespace declarations that should be skipped.
+     */
+    const std::set<const clang::NamespaceDecl*>& GetNamespacesToSkip() const;
 
     /**
      * Get the YAML configuration associated with a specific declaration,
@@ -219,7 +222,7 @@ protected:
     std::vector<std::pair<const clang::QualType, YAML::Node>> types_;
     std::map<const clang::Decl*, YAML::Node> declarations_;
     std::set<const clang::NamespaceDecl*> namespaces_;
-    std::set<const clang::NamespaceDecl*> namespaceBlackList_;
+    std::set<const clang::NamespaceDecl*> namespacesToSkip_;
 
     std::vector<std::string> binding_names_;
     std::vector<std::shared_ptr<chimera::mstch::Namespace>> binding_namespaces_;
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -141,6 +141,8 @@ public:
      */
     const std::set<const clang::NamespaceDecl*>& GetNamespaces() const;
 
+    const std::set<const clang::NamespaceDecl*>& GetNamespaceBlacklist() const;
+
     /**
      * Get the YAML configuration associated with a specific declaration,
      * or return an empty YAML node if no configuration was found.
@@ -217,6 +219,7 @@ protected:
     std::vector<std::pair<const clang::QualType, YAML::Node>> types_;
     std::map<const clang::Decl*, YAML::Node> declarations_;
     std::set<const clang::NamespaceDecl*> namespaces_;
+    std::set<const clang::NamespaceDecl*> namespaceBlackList_;
 
     std::vector<std::string> binding_names_;
     std::vector<std::shared_ptr<chimera::mstch::Namespace>> binding_namespaces_;
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -221,7 +221,7 @@ protected:
     clang::CompilerInstance *ci_;
     std::vector<std::pair<const clang::QualType, YAML::Node>> types_;
     std::map<const clang::Decl*, YAML::Node> declarations_;
-    std::set<const clang::NamespaceDecl*> namespaces_;
+     std::set<const clang::NamespaceDecl*> namespacesIncluded_;
     std::set<const clang::NamespaceDecl*> namespacesToSkip_;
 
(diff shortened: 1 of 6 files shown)


jslee02:
@psigen Please review this PR when you have a chance. This change really helps for building dartpy without any manual changes required from the auto-generated bindings.

include/chimera/configuration.h
```diff
@@ -217,6 +222,7 @@ class CompiledConfiguration
     std::vector<std::pair<const clang::QualType, YAML::Node>> types_;
     std::map<const clang::Decl*, YAML::Node> declarations_;
     std::set<const clang::NamespaceDecl*> namespaces_;
```
psigen:
```suggestion
     std::set<const clang::NamespaceDecl*> namespacesIncluded_;
```

include/chimera/configuration.h
```diff
@@ -217,6 +222,7 @@ class CompiledConfiguration
...
+    std::set<const clang::NamespaceDecl*> namespacesToSkip_;
```
psigen:
```suggestion
    std::set<const clang::NamespaceDecl*> namespacesSuppressed_;
```

include/chimera/configuration.h
```diff
@@ -141,6 +141,11 @@ class CompiledConfiguration
      */
     const std::set<const clang::NamespaceDecl*>& GetNamespaces() const;
```
psigen:
```suggestion
    const std::set<const clang::NamespaceDecl*>& GetNamespacesIncluded() const;
```

include/chimera/configuration.h
```diff
@@ -141,6 +141,11 @@ class CompiledConfiguration
...
 
+    /**
+     * Returns list of namespace declarations that should be skipped.
+     */
+    const std::set<const clang::NamespaceDecl*>& GetNamespacesToSkip() const;
```
psigen:
```suggestion
    const std::set<const clang::NamespaceDecl*>& GetNamespacesSuppressed() const;
```

src/configuration.cpp
```diff
@@ -206,14 +206,21 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
             {
                 std::string ns_str = it.first.as<std::string>();
                 auto ns = chimera::util::resolveNamespace(ci, ns_str);
                 if (ns)
                 {
-                    declarations_[ns] = it.second;
-                    namespaces_.insert(ns);
+                    if (it.second.IsNull())
+                    {
+                        namespacesToSkip_.insert(ns);
```
psigen:
```suggestion
                        namespacesSuppressed_.insert(ns);
```

src/configuration.cpp
```diff
@@ -206,14 +206,21 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+                    }
+                    else
+                    {
+                        declarations_[ns] = it.second;
+                        namespaces_.insert(ns);
```
psigen:
```suggestion
                        namespacesIncluded_.insert(ns);
```

src/configuration.cpp
```diff
@@ -461,6 +468,13 @@ chimera::CompiledConfiguration::~CompiledConfiguration()
 void
 chimera::CompiledConfiguration::AddTraversedNamespace(const clang::NamespaceDecl* decl)
 {
+    // Skip namespaces that are defined as null in the configuration.
+    for (const auto &it : GetNamespacesToSkip())
```
psigen:
```suggestion
    for (const auto &it : GetNamespacesSuppressed())
```

src/configuration.cpp
```diff
@@ -481,6 +495,12 @@ chimera::CompiledConfiguration::GetNamespaces() const
     return namespaces_;
 }
 
+const std::set<const clang::NamespaceDecl*>&
+chimera::CompiledConfiguration::GetNamespacesToSkip() const
```
psigen:
```suggestion
chimera::CompiledConfiguration::GetNamespacesSuppressed() const
```

src/configuration.cpp
```diff
@@ -481,6 +495,12 @@ chimera::CompiledConfiguration::GetNamespaces() const
...
+{
+    return namespacesToSkip_;
```
psigen:
```suggestion
    return namespacesSuppressed_;
```

src/configuration.cpp
```diff
@@ -481,6 +495,12 @@ chimera::CompiledConfiguration::GetNamespaces() const
     return namespaces_;
```
psigen:
```suggestion
     return namespacesIncluded_;
```

src/configuration.cpp
```diff
@@ -512,6 +532,13 @@ clang::ASTContext &chimera::CompiledConfiguration::GetContext() const
 
 bool chimera::CompiledConfiguration::IsEnclosed(const clang::Decl *decl) const
 {
+    // Skip namespaces that are defined as null in the configuration.
+    for (const auto &it : GetNamespacesToSkip())
```
psigen:
```suggestion
    for (const auto &it : GetNamespacesSuppressed())
```

src/configuration.cpp
```diff
@@ -512,6 +532,13 @@ clang::ASTContext &chimera::CompiledConfiguration::GetContext() const
...
+    {
+        if (decl->getDeclContext() && it->Encloses(decl->getDeclContext()))
+            return false;
+    }
+
     // Filter over the namespaces and only traverse ones that are enclosed
     // by one of the configuration namespaces.
     for (const auto &it : GetNamespaces())
```
psigen:
```suggestion
     for (const auto &it : GetNamespacesIncluded())
```

jslee02:
> It looks like we might be able to put this logic into the top of the `IsSuppressed(decl)` function, and then use that in the position of `IsEnclosed()`.

I like the idea that making `IsSuppressed()` to check suppressed namespace. But I'm not sure if you mean removing `IsEnclosed()` in favor of `IsSuppressed()`. Could you let me know what was your intention?

psigen:
Sorry: here is what I think makes sense:
* Put the checks against both `GetNamespaceSuppressed()` and `GetNamespaceIncluded()` into `IsEnclosed()`, as it was before.
* Make `if (!IsEnclosed(decl)) { return true; }` be the first check of `IsSuppressed()` so we don't repeat the namespace enclosure checking logic.
* Call `IsSuppressed()` instead of `IsEnclosed()` in [visitor.cpp](https://github.com/personalrobotics/chimera/blob/master/src/visitor.cpp#L98).
* Remove other usages of `IsSuppressed()` (because I _think_ they will be superseded by the one initial call at `VisitDecl`).  **This is the step that might introduce issues.**

Does that make sense?  The only thing I am unsure of is whether rearranging these checks will have any weird side-effects.  I don't see why they should but I'm not sure.

jslee02:
Thanks for the guide! I tested this on `dartpy`, and it worked as expected without introducing any side effects. Let me merge this!

PR #239, opened 2019-01-25
Strip non-copyable Eigen wrapper classes for Pybind11

jslee02:
Suppose we have a template function with Eigen base class such as:
```cpp
// https://eigen.tuxfamily.org/dox/TopicFunctionTakingEigenTypes.html
template <typename Derived>
void print_size(const EigenBase<Derived>& b)
{
  std::cout << "size (rows, cols): " << b.size() << " (" << b.rows()
            << ", " << b.cols() << ")" << std::endl;
}
```
Chimera generates binding for the function something like (for Pybind11):
```cpp
void binding(pybind11::module& m)
{
    m.def("print_size", +[](const Eigen::EigenBase<Eigen::Matrix<double, 3, 3, 0, 3, 3> > & b) { chimera_test::print_size<Eigen::EigenBase<Eigen::Matrix<double, 3, 3, 0, 3, 3>>>(b); }, ::pybind11::arg("b"));
}
```

However, this is not compilable with Pybind11 because [Pybind11 only supports Eigen types that are default-constructible](https://github.com/pybind/pybind11/issues/1324#issuecomment-374031928) and `EigenBase<...>` is not default-constructible.

This PR adds a function `stripNoneCopyableEigenWrappers()` (and some necessary changes to support the function) to extract the derived class so that we the bindings become:
```cpp
void binding(pybind11::module& m)
{
    m.def("print_size", +[](const Eigen::Matrix<double, 3, 3, 0, 3, 3> & b) { chimera_test::print_size<Eigen::Matrix<double, 3, 3, 0, 3, 3>>(b); }, ::pybind11::arg("b"));
}
```
which is compilable.


jslee02:
I admit that hard-coding for a specific binder and a library is not the best way to resolve the issue. Also, I think manipulating strings instead of the clang types isn't a good way. But this would be a stopgap solution to move on. I am definitely open to better solutions, but highly prefer to have a working version for our use case (i.e., dartpy).

> Is there perhaps a more general way to describe what is happening here not in terms of a specific quirk of pybind11?

This change is not about a bug of Chimera, but just Pybind11 specific workaround.

> Perhaps what is really missing is a configuration to suppress output of a specific list of template wrappers by name or regex, and one such use case is the list of Eigen classes, which should be specified in a configuration file.

This is not just for suppressing the specific list of templates but also generate compilable code from the templates. I'm not sure if the current (or small modification of) configuration would handle this.

jslee02:
Created #250 as a proposal to improve the changes in this PR.

Let me merge this for now.

PR #246, opened 2019-01-29
Build with LLVM version equal to or greater than system version

jslee02:
* Trusty: Use 3.6. The system version is 3.4, but the minimum required version is 3.6.
* Xenial: Use 3.9. The system version is 3.8, but there is no available llvm-tool 3.8.
* Bionic: Use 6.0, which is the system version
* Cosmic: Use 6.0, which is the system version


jslee02:
Travis and Azure install `llvm` to different locations. `script.sh` should be able to find the locations accordingly.

PR #251, opened 2019-01-30
Fix going to wrong directory while running rebuild target

jslee02:
Otherwise, this causes CMake error in downstream projects as:
```shell
[100%] Performing build step for 'dartpy_CHIMERA_REBUILD'
[100%] No install step for 'dartpy_CHIMERA_REBUILD'
[100%] Completed 'dartpy_CHIMERA_REBUILD'
[100%] Built target dartpy_CHIMERA_REBUILD
python/dartpy/CMakeFiles/binding.dir/build.make:54: CMakeFiles/binding.dir/progress.make: No such file or directory
make[3]: *** No rule to make target 'CMakeFiles/binding.dir/progress.make'.  Stop.
CMakeFiles/Makefile2:6246: recipe for target 'python/dartpy/CMakeFiles/binding.dir/all' failed
make[2]: *** [python/dartpy/CMakeFiles/binding.dir/all] Error 2
CMakeFiles/Makefile2:6258: recipe for target 'python/dartpy/CMakeFiles/binding.dir/rule' failed
make[1]: *** [python/dartpy/CMakeFiles/binding.dir/rule] Error 2
Makefile:1761: recipe for target 'binding' failed
make: *** [binding] Error 2

```

@psigen Was there a reason of using `ExternalProject_Add()` instead of `add_custom_target()`?

### DIFF
--- a/cmake/chimeraFunctions.cmake
+++ b/cmake/chimeraFunctions.cmake
@@ -182,9 +182,10 @@ function(add_chimera_binding)
         DOWNLOAD_COMMAND ""
         INSTALL_COMMAND ""
         BUILD_COMMAND $(MAKE) "${binding_TARGET}_SOURCES"
+        COMMAND cd "${PROJECT_BINARY_DIR}"
         DEPENDS "${binding_TARGET}_SOURCES"
         SOURCE_DIR "${PROJECT_SOURCE_DIR}"
-        BUILD_IN_SOURCE 1
+        BINARY_DIR "${PROJECT_BINARY_DIR}"
     )
     set_target_properties("${binding_TARGET}_REBUILD" PROPERTIES EXCLUDE_FROM_ALL TRUE)
     add_dependencies("${binding_TARGET}" "${binding_TARGET}_REBUILD")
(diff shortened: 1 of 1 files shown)


psigen:
`add_custom_target()` did not seem sufficient to rebuild files correctly,
because it does not have the capability to re-run the binding generation
(another cmake step).

We have a dependency loop where the binding sources are being generated by
a cmake step and then consumed by another cmake step (the sources list of
the output built target). I used `ExternalProject_Add` to break this
dependency loop, because it will run the internal CMake target separately,
meaning that it can change the binding sources and be appropriately re-run
without breaking the caching around the sources list.

I strongly suspect that if you make this change, it will lead to transient
build errors where the target fails on the first build and succeeds on the
second build, and there will be no way to solve that without this technique.

jslee02:
Thanks for the quick replay! That makes sense (and the tests are failing due to this change). Do you have any idea then the cause of errors in the post?

The command sequence I did:
```
$ cd dartpy
$ mkdir build; cd build
$ cmake ..
$ make binding // <-- this causes the above errors
$ cmake ..
$ make binding // Now no error!
```

jslee02:
I think I found the right solution! Using source directory for building directory resolves this. This is done by adding [`BUILD_IN_SOURCE` option](https://cmake.org/cmake/help/v3.9/module/ExternalProject.html), which is equivalent to `BINARY_DIR "${PROJECT_SOURCE_DIR}"`.

PR #259, opened 2019-02-19
Fix unnecessarily requiring (nested) namespace for Boost.Python bindings

jslee02:
### Motivation

Chimera cannot correctly generates `boost_python` bindings for objects that are not in a (nested) namespace. This is because the `boost_python` template generates following bindings for the cases:
```cpp
    ::boost::python::object parent_object(::boost::python::scope());  // invalid!
```
The compiler complains as:
```
chimera/build/test/examples/01_function/function_boost_python/_ZN12chimera_test3addEii.cpp: In function ‘void _ZN12chimera_test3addEii()’:
chimera/build/test/examples/01_function/function_boost_python/_ZN12chimera_test3addEii.cpp:11:54: error: no matching function for call to ‘boost::python::scope::scope(boost::python::api::object (&)(boost::python::scope (*)()))’
                                                      ^
In file included from /usr/include/boost/python/def.hpp:14,
                 from /usr/include/boost/python.hpp:22,
                 from /home/js/dev/prl/chimera/build/test/examples/01_function/function_boost_python/_ZN12chimera_test3addEii.cpp:4:
/usr/include/boost/python/scope.hpp:44:8: note: candidate: ‘boost::python::scope::scope()’
 inline scope::scope()
        ^~~~~
/usr/include/boost/python/scope.hpp:44:8: note:   candidate expects 0 arguments, 1 provided
/usr/include/boost/python/scope.hpp:37:8: note: candidate: ‘boost::python::scope::scope(const boost::python::api::object&)’
 inline scope::scope(object const& new_scope)
        ^~~~~
/usr/include/boost/python/scope.hpp:37:8: note:   no known conversion for argument 1 from ‘boost::python::api::object(boost::python::scope (*)())’ to ‘const boost::python::api::object&’
/usr/include/boost/python/scope.hpp:68:8: note: candidate: ‘boost::python::scope::scope(const boost::python::scope&)’
 inline scope::scope(scope const& new_scope)
        ^~~~~
/usr/include/boost/python/scope.hpp:68:8: note:   no known conversion for argument 1 from ‘boost::python::api::object(boost::python::scope (*)())’ to ‘const boost::python::scope&’

```
For this reason, the binding tests have to have `nested_namespace` so that the binding becomes:
```cpp
    ::boost::python::object parent_object(::boost::python::scope().attr("nested_namespace"));  // okay
```

Note that we intentionally collapse the root namespace `chimera_test` by setthing it to null to avoid repeating the scope name. Having the root namespace, the binding could be successfully compiled, and the python code would be like:
```python
import chimera_test

chimera_test.chimera_test.add(3, 4)
```
The first `chimera_test` is from the module name, and the second `chimera_test` from the root namespace. However, we generally collapse the root namespace to avoid this undesired case.

### Solution

The `boost_python` template are modified to generate the scope code lines only when the binding object has at least one _named_ scope item. To support the modified template, following changes were required in the C++ code:

* Add emptiness checking mstsc methods for the scopes: `scope?`, `namespaceScope?`, and `classScope?`
* Add `ClangWrapper::nameAsString()` to be able to check emptiness of the mstsc object
* Exclude nonamed namespace and class from the scopes, which I believe this doesn't harm the other functionalities because we don't use them anyways.

Each scope item should have name. Otherwise, it could still generate the same invalid binding.


psigen:
Thanks @jslee02 , but maybe we can also keep some unit tests for a nested namespace to make sure that still works as well?

jslee02:
Oh, I thought we have one [here](https://github.com/personalrobotics/chimera/blob/master/test/examples/02_class/class.py#L56), but it actually tests of ignoring namespace. Let me add one.

Issue #228, opened 2018-10-26
Incorrect bindings for type alias in different namespace from underlying type


jslee02:
For the following C++ code
```cpp
namespace chimera_test { namespace nested_ns {
template <class T> using custom_shared_ptr = std::shared_ptr<T>;
void take_custom_shared_ptr(custom_shared_ptr<int> val) {}
} }
```
, chimera generates the python binding code as
```cpp
void _ZN8function15nested_function22take_custom_shared_ptrESt10shared_ptrIiE()
{
::boost::python::object parent_object(::boost::python::scope().attr("nested_function"));
::boost::python::scope parent_scope(parent_object);

::boost::python::def("take_custom_shared_ptr", +[](std::custom_shared_ptr<int> val) { function::nested_function::take_custom_shared_ptr(val); }, (::boost::python::arg("val")));

}
```
which causes build error due to `std::custom_shared_ptr<int>`.

Note that `std::custom_shared_ptr<int>` should be `chimera_test::nested_ns::custom_shared_ptr<int>`.

jslee02:
Update: This happens when the type alias is declared in a different namespace from the underlying type. Function `GetFullyQualifiedType(...)` in file `cling_utils_AST.h` always adds the namespace of the underlying type to the type alias, which should be the namespace of the local namespace the type alias is declared. Note that this has nothing to do with STL headers nor template.

Issue #218, opened 2018-09-25
Test fails on macOS: fails to include STL headers


jslee02:
https://travis-ci.org/personalrobotics/chimera/jobs/432802849#L1659

PR #263, opened 2020-04-03
Update CI reflecting recent OS updates

jslee02:
- Update CI images to recent OS versions
- Remove CI builds for deprecated OS versions
- Add CI builds for new OS versions
- Disable building tests for macOS (see: [#228](https://github.com/personalrobotics/chimera/issues/218)) and Ubuntu Focal (tests  fails for some reason)
- Install `yaml-cpp` without option `--with-static-lib` as it's removed
- Update `README.md` reflecting the updates

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -4,7 +4,7 @@
 jobs:
 - job: ubuntu_xenial_gcc_release
   pool:
-    vmImage: 'ubuntu-16.04'
+    vmImage: 'ubuntu-18.04'
   variables:
     OS_NAME: linux
     COMPILER: gcc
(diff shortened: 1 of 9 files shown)


jslee02:
> Do you think we should switch over to using GitHub actions for the CI? I have been having good success with it on other projects.

Yeah, DART also uses GitHub Actions along other CIs. DART still uses other CIs to run the CI builds in parallel as many as possible with different OSs and configurations. For Chimera, one CI would be sufficient I think, and GitHub actions seems to be a good choice as it provides all the features that other CI services provide.

Another motivation DART is still using Travis is to push/deploy the generated API documents and coverage reports. But this is just because I didn't check if GitHub actions also supports this feature.

jslee02:
Oh, wait. It seems I'm no longer authorized to merge PRs of Chimera. :cry:

@egordon Could you add me as a contributor of this repository?

psigen:
Me either...

jslee02:
> Me either...

We should be able to merge PRs now thanks to @egordon!

PR #264, opened 2020-04-04
Workaround for segfault with LLVM >= 7

jslee02:
I wasn't able to find the root cause, but build with LLVM >= 7 causes segfault when the destructor of `clang::Parser` is called in `chimera::util::resolveDeclaration`.

This PR workaround by not destructing `clang::Parser` for LLVM >= 7. Since this workaround leads to a memory leak, this should be fixed once we figure out the root cause.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -116,10 +116,17 @@ const NamedDecl *resolveDeclaration(CompilerInstance *ci,
     // Create a new parser to handle this type parsing.
     Preprocessor &preprocessor = ci->getPreprocessor();
     Sema &sema = ci->getSema();
+#if LLVM_VERSION_AT_LEAST(7, 0, 0)
+    // TODO: Since LLVM 7, segfault occurs when the destructor of clang::Parser
+    // is called. Following code workaround the segfault by not destructing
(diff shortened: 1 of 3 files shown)


src/util.cpp
```diff
@@ -116,7 +116,17 @@ const NamedDecl *resolveDeclaration(CompilerInstance *ci,
     // Create a new parser to handle this type parsing.
     Preprocessor &preprocessor = ci->getPreprocessor();
     Sema &sema = ci->getSema();
+#if LLVM_VERSION_AT_LEAST(7, 0, 0)
+    // TODO: Since LLVM 7, segfault occurs when the destructor of clang::Parser
+    // is called. Following code workaround the segfault by not destructing
+    // the clang::Parser instance, which leads to memory leak. This should be
+    // fixed. See https://github.com/personalrobotics/chimera/issues/222
```
psigen, line 123: I wonder if this could be related:
https://github.com/Papierkorb/bindgen/pull/20

jslee02: Thanks for the link, but it also simply disables the code that causes the segfault (i.e., `clang::ParseAST` in that case) without fixing the root problem. :disappointed:

PR #267, opened 2020-04-04
Change exit() instances to throw exceptions

### LINKED ISSUE #136: Change `exit()` instances to throw exceptions.
Currently, we `exit()` in various places in chimera where we encounter unexpected states.  Instead, in #135 @mkoval suggested we switch to throwing exceptions in those places.

jslee02:
Fixes #136

### DIFF
--- a/src/configuration.cpp
+++ b/src/configuration.cpp
@@ -2,6 +2,7 @@
 #include "chimera/mstch.h"
 #include "chimera/util.h"
 
+#include <exception>
 #include <fstream>
 #include <iomanip>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


src/configuration.cpp
```diff
@@ -408,7 +400,7 @@ chimera::CompiledConfiguration::CompiledConfiguration(
         = [](const std::string &str) -> std::string { return str; };
 }
 
-chimera::CompiledConfiguration::~CompiledConfiguration()
+chimera::CompiledConfiguration::~CompiledConfiguration() noexcept(false)
```
psigen, line 403: It is a bit weird to be doing this in the destructor.  I wonder if there is a different location in the visitor call chain where this would make more sense :thinking:

jslee02:
All sounds good. One issue would be that `Visitor` takes `std::unique_ptr<CompiledConfiguration>` so we have to move the created unique_ptr to the constructor of `Visitor` and then cannot access to the compiled config.

Possible solutions would be:
1. change the constructor of `Visitor` to take `shared_ptr<CompiledConfiguration>`
1. add a method to `Visitor` to return a raw pointer to the compiled config
Which one would be good for this case?

jslee02:
Merging as the comment is resolved by #269

PR #269, opened 2020-04-05
Create a dedicated function to render top-level mstch

jslee02:
A possible solution to this [comment](https://github.com/personalrobotics/chimera/pull/267#pullrequestreview-387768668).

### DIFF
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -217,8 +217,8 @@ public:
     bool Render(const std::shared_ptr<chimera::mstch::Variable> context);
 
     /**
-     * Renders the top-level mstch template where the filename is specified by
-     * Configuration::SetOutputModuleName().
+     * Renders the top-level mstch template. The rendered filename is specified
+     * by Configuration::SetOutputModuleName().
      */
     void Render();
 
--- a/include/chimera/configuration.h
+++ b/include/chimera/configuration.h
@@ -121,7 +121,7 @@ protected:
 class CompiledConfiguration
 {
 public:
-    virtual ~CompiledConfiguration();
+    virtual ~CompiledConfiguration() = default;
     CompiledConfiguration(const CompiledConfiguration &) = delete;
     CompiledConfiguration &operator=(const CompiledConfiguration &) = delete;
 
@@ -215,6 +215,12 @@ public:
     bool Render(const std::shared_ptr<chimera::mstch::Enum> context);
     bool Render(const std::shared_ptr<chimera::mstch::Function> context);
     bool Render(const std::shared_ptr<chimera::mstch::Variable> context);
+
+    /**
+     * Renders the top-level mstch template where the filename is specified by
+     * Configuration::SetOutputModuleName().
+     */
+    void Render();
 
 private:
     CompiledConfiguration(const Configuration &parent,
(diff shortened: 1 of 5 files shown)


src/consumer.cpp
```diff
@@ -15,9 +15,15 @@ void chimera::Consumer::HandleTranslationUnit(ASTContext &context)
 {
     // Use the current translation unit to resolve the YAML configuration.
     chimera::Configuration &config = chimera::Configuration::GetInstance();
-    chimera::Visitor visitor(ci_, config.Process(ci_));
+
+    std::shared_ptr<chimera::CompiledConfiguration> compiled_config
+        = config.Process(ci_);
```
psigen, line 20: Does this even need to be a pointer anymore?  We can just pass by reference into the visitor, because we know it will stay in scope over the lifetime of the visitor, right?
```suggestion
chimera::CompiledConfiguration compiled_config = config.Process(ci_);
```

include/chimera/configuration.h
```diff
@@ -216,6 +216,12 @@ class CompiledConfiguration
     bool Render(const std::shared_ptr<chimera::mstch::Function> context);
     bool Render(const std::shared_ptr<chimera::mstch::Variable> context);
 
+    /**
+     * Renders the top-level mstch template where the filename is specified by
+     * Configuration::SetOutputModuleName().
+     */
+    void Render();
```
psigen, line 223: Hmm, I see that there are other functions called `Render()`, we should probably pick a different name here.
```suggestion
    void Serialize();
```

include/chimera/visitor.h
```diff
@@ -15,7 +16,7 @@ class Visitor : public clang::RecursiveASTVisitor<Visitor>
 {
 public:
     Visitor(clang::CompilerInstance *ci,
-            std::unique_ptr<CompiledConfiguration> cc);
+            std::shared_ptr<CompiledConfiguration> cc);
```
psigen, line 19: Since we no longer need ownership at all and know the configuration will stay in scope, can we pass by reference here?
```suggestion
            CompiledConfiguration &cc);
```

src/consumer.cpp
jslee02: The scope seems okay, but in order to make `Configuration::Process()` to return a value of `CompiledConfiguration`, we have to un-delete copy constructor (and/or move constructor) of `CompiledConfiguration`, which is currently deleted.

We could either (1) implement the necessary methods, (2) create `unique_ptr`  and pass dereferenced value, or (3) use the current implementation in this PR. Which one would you prefer?

include/chimera/visitor.h
jslee02: This decision depends on the first comment.

include/chimera/configuration.h
jslee02: I considered that but didn't because the other Render() functions (including the protected one) indeed have similar purposes. The only difference is that this `Render()` generates the top-level bindings while others generate sub bindings for variables, functions, class, etc.

I would be fine to rename this to `Serialize()` as well.

psigen: Ok, we can leave it alone then, this seems fine.

src/consumer.cpp
psigen: Let's do (2), that is the most similar to the existing behavior.

jslee02: Done in https://github.com/personalrobotics/chimera/pull/269/commits/d68eceeb3cab3d1efeee3356faefa7f7e2c1b547

PR #270, opened 2020-04-05
Update build instruction in README.md

jslee02:


README.md
```diff
@@ -95,8 +102,8 @@ $ sudo make install
 ### On macOS from Source
 
 ```bash
-$ brew install boost llvm
-$ brew install yaml-cpp --with-static-lib
```
psigen, line 99: Do we no longer need the static lib here?

jslee02: I believe it was optional so that chimera doesn't require the shared library of `yaml-cpp`. The actual reason of this change is that `--with-static-lib` option is removed from the Homebrew formula of `yaml-cpp` so no longer able to use the option. I'm not sure if the default build option of `yaml-cpp` is now static or shared though.

psigen: Ok, we might want to just check that this is all still working, but if the unit tests are still working on OSX it is probably fine.

jslee02: Agreed. However, it's been a while since we don't run unit tests on macOS due to https://github.com/personalrobotics/chimera/issues/218. :disappointed:

Removing the option was already made in https://github.com/personalrobotics/chimera/pull/263. This PR just reflects the change.

PR #281, opened 2020-04-12
Eliminate singleton configuration with custom FrontendActionFactory.

psigen:
This PR eliminates the need for the singleton instantiation of the `chimera::Configuration` by implementing custom factory classes that allow the configuration to be passed to the ASTConsumer directly.


CMakeLists.txt
```diff
@@ -95,9 +95,9 @@ set(${PROJECT_NAME}_SOURCES
   src/configuration.cpp
   src/consumer.cpp
   src/frontend_action.cpp
-  src/visitor.cpp
-  src/util.cpp
   src/mstch.cpp
+  src/util.cpp
+  src/visitor.cpp
```
psigen, line 100: I just reordered these to be in alphabetical order.

CMakeLists.txt
```diff
@@ -171,8 +171,11 @@ install(FILES "${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
 )
 
 # Enable unit testing by including the `test` subdirectory.
-enable_testing()
-add_subdirectory(test EXCLUDE_FROM_ALL)
+option(BUILD_TESTING "Enable building tests" ON)
+if(BUILD_TESTING)
+  enable_testing()
+  add_subdirectory(test EXCLUDE_FROM_ALL)
+endif()
```
psigen, line 178: I saw that there was a `BUILD_TESTING` flag being used, but it didn't seem to affect the build.  I needed to build without tests (I didn't have python 3.7 installed), so I added this option.

Feel free to suggest a different name for the flag.

Dockerfile
```diff
@@ -1,14 +1,15 @@
-FROM ubuntu:16.04 as builder
+FROM ubuntu:18.04 as builder
```
psigen, line 1: I updated the contents of this Dockerfile to successfully build against llvm-6.0 / ubuntu 18.04.

include/chimera/frontend_action.h
```diff
@@ -14,10 +19,39 @@ namespace chimera
...
+protected:
+    const chimera::Configuration &config_;
 };
 
+/**
+ * Custom frontend factory that forwards a ChimeraConfiguration.
+ */
+std::unique_ptr<clang::tooling::FrontendActionFactory> newFrontendActionFactory(
+    const chimera::Configuration &config);
+
```
psigen, line 54: This approach is summarized in this comment thread requesting this functionality be added to LLVM.  It doesn't look like it was added though :(

http://lists.llvm.org/pipermail/cfe-dev/2014-December/040607.html

src/frontend_action.cpp
```diff
@@ -17,5 +23,34 @@ std::unique_ptr<clang::ASTConsumer> chimera::FrontendAction::CreateASTConsumer(
...
+
+// Create a custom action factory that forwards the ChimeraConfiguration.
+chimera::ChimeraFrontendActionFactory::ChimeraFrontendActionFactory(
+    const chimera::Configuration &config)
+  : config_(config)
+{
+    // Do nothing.
+}
+
+#if LLVM_VERSION_AT_LEAST(8, 0, 0)
```
psigen, line 38: This version code might be slightly wrong.  Somewhere between llvm-6 and llvm-10, this interface changed from returning a raw pointer to returning a `std::unique_ptr`.

CMakeLists.txt
jslee02: [`BUILD_TESTING` is automatically set when we include `CTest`](https://cmake.org/cmake/help/v3.7/module/CTest.html). I think the issue is that `CTest` is being included in `test/examples/CMakeLists.txt` (line: 1). Let's replace line 174 with `include(CTest)` and remove line 1 of `test/examples/CMakeLists.txt`.

psigen: Does this look like what you had in mind?

psigen:
Thanks @jslee02, made the suggested changes in https://github.com/personalrobotics/chimera/pull/281/commits/0ddac1b97d7e91d9fe663fd4bca91e5225c86939

psigen:
Ok, I think I made all the changes now.

PR #282, opened 2020-04-12
Fix build errors for #281

jslee02:
This PR fixes build errors for #281. Note that the target branch is `no_single_confing`.


include/chimera/frontend_action.h
```diff
@@ -42,7 +42,7 @@ class ChimeraFrontendActionFactory
 // FrontendActionFactory::create() changed from a raw pointer to a
 // std::unique_ptr, but we haven't specifically checked which version
 // the change occurred in yet.
-#if LLVM_VERSION_AT_LEAST(8, 0, 0)
+#if LLVM_VERSION_AT_LEAST(9, 0, 0)
```
jslee02, line 45: I tested with LLVM 8, but still the return type is `clang::FrontendAction *`.

jslee02: It seems LLVM 9 is the same. See https://dev.azure.com/personalrobotics/chimera/_build/results?buildId=486&view=logs&j=44919602-b7ab-5713-e632-82bce96922d8&t=431ee554-ac75-5d0b-a937-0313a76e2825&l=116

PR #284, opened 2020-04-12
Add docstring to function templates

jslee02:


jslee02:
> I'm amazed that this is working with so little modification

Yeah, I think it's the beauty of Chimera. :sunglasses:

Issue #288, opened 2020-04-13
Consider splitting bindings into header and source


jslee02:
Header bindings are required for the case that bindings need to include other bindings to compile. For example, "trampoline" classes in Python need to include the trampoline classes of the parent classes, which are currently generated in the source bindings. See `PyAnimal` and `PyDog` example in this [link](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#combining-virtual-functions-and-inheritance). `PyDog` should inherit `PyAnimal`, but it cannot because `PyAnimal` is generated in a separate source file.

A possible solution would be having header and source templates for each binding type (e.g., class, enum, function, and variable; maybe except for module?).

jslee02:
Addressed by #289

PR #289, opened 2020-04-13
Split bindings into header and source

jslee02:
A quick prototype implementation for #288.

Binding templates are split into headers and sources except for the module templates.

This change breaks the backward compatibility in the parsing template in the configuration because the new template configuration requires two files for header and source templates for each type as:
```yaml
# OLD
template:
  module: !file
    'templates/module.mstch.cpp'
  class: !file
    'templates/cxx_record.mstch.cpp'
  enum: !file
    'templates/enum.mstch.cpp'
  function: !file
    'templates/function.mstch.cpp'
  variable: !file
    'templates/variable.mstch.cpp'
```
```yaml
# NEW
template:
  module: !file
    'templates/module.mstch.cpp'
  class:
    header: !file
      'templates/cxx_record.mstch.cpp'
    source: !file
      'templates/cxx_record.mstch.h'
  enum:
    header: !file
      'templates/enum.mstch.cpp'
    source: !file
      'templates/enum.mstch.h'
  function:
    header: !file
      'templates/function.mstch.cpp'
    source: !file
      'templates/function.mstch.h'
  variable:
    header: !file
      'templates/variable.mstch.cpp'
    source: !file
      'templates/variable.mstch.h'
```
We could make this backward compatible if desired.

We may also need to change tags under `template.file` to be able to set separately for header and source.


src/configuration.cpp
```diff
@@ -451,20 +453,36 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
-        if (const YAML::Node functionTemplateNode = bindingNode_["function"])
-            bindingDefinition_.function_cpp = Lookup(functionTemplateNode);
+        if (const auto node
+            = lookupYAMLNode(bindingNode_, "function", "source"))
+            bindingDefinition_.function_cpp = Lookup(node);
 
-        if (const YAML::Node moduleTemplateNode = bindingNode_["module"])
-            bindingDefinition_.module_cpp = Lookup(moduleTemplateNode);
+        if (const auto node = bindingNode_["module"])
+            bindingDefinition_.module_cpp = Lookup(node);
```
psigen, line 477: Let's use `lookupYAMLNode()` here as well for consistency.
```suggestion
        if (const auto node = lookupYAMLNode(bindingNode_, "module"))
            bindingDefinition_.module_cpp = Lookup(node);
```

src/configuration.cpp
```diff
@@ -718,7 +736,9 @@ std::string chimera::CompiledConfiguration::Lookup(const YAML::Node &node) const
 }
 
 bool chimera::CompiledConfiguration::Render(
-    std::string view, std::string key,
+    const std::string &key, const std::string &header_view,
+    const std::string &source_view, const std::string &header_ext,
+    const std::string &source_ext,
```
psigen, line 741: Why do we pass in the `header_ext` and the `source_ext` if they are always the same values?  Can't we just specify them in this function when we call the other `Render()`?

jslee02: It makes sense to me. Done in https://github.com/personalrobotics/chimera/pull/289/commits/9f8d9b5a78d3cfa31640af349fcec377e09238d7.

src/configuration.cpp
```diff
@@ -451,20 +453,36 @@ chimera::CompiledConfiguration::CompiledConfiguration(
     // Override individual templates if specified in the configuration.
     if (bindingNode_)
     {
-        if (const YAML::Node classTemplateNode = bindingNode_["class"])
-            bindingDefinition_.class_cpp = Lookup(classTemplateNode);
+        if (const auto node = lookupYAMLNode(bindingNode_, "class", "header"))
+            bindingDefinition_.class_h = Lookup(node);
+
+        if (const auto node = lookupYAMLNode(bindingNode_, "class", "source"))
+            bindingDefinition_.class_cpp = Lookup(node);
```
psigen, line 460: Hmm, I just had a thought.
What if we do the following to support both types?
```suggestion
        if (const auto node = bindingNode_["class"]) {
            if (node.IsScalar()) {
                bindingDefinition_.class_cpp = Lookup(node);
            } else {
                bindingDefinition_.class_h = Lookup(node["header"]);
                bindingDefinition_.class_cpp = Lookup(node["source"]);
            }            
        }
```
Does it work?

src/configuration.cpp
```diff
@@ -451,20 +453,36 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
-        if (const YAML::Node functionTemplateNode = bindingNode_["function"])
-            bindingDefinition_.function_cpp = Lookup(functionTemplateNode);
+        if (const auto node
+            = lookupYAMLNode(bindingNode_, "function", "source"))
+            bindingDefinition_.function_cpp = Lookup(node);
 
-        if (const YAML::Node moduleTemplateNode = bindingNode_["module"])
-            bindingDefinition_.module_cpp = Lookup(moduleTemplateNode);
+        if (const auto node = lookupYAMLNode(bindingNode_, "module"))
+            bindingDefinition_.module_cpp = Lookup(node);
```
psigen, line 477: Should we allow header and source on module just for consistency?  It is a bit abnormal but maybe it's ok?

jslee02: Sounds fair to me. Done in https://github.com/personalrobotics/chimera/pull/289/commits/468e6aa490e4aa1292e01e876cd052a4e009bc2f.

src/configuration.cpp
jslee02: Yes, this would work (done in https://github.com/personalrobotics/chimera/pull/289/commits/6c66716240bdf16ed99e21ac29cc725659d0dec0).

My only concern, however, is the case that the user only provides a source template using the old syntax and expect no header template is being generated. In this case, the user-provided source template would be likely not compatible with the default header template of Chimera.

If we really want backward compatibility, then we need to add a logic not to generate headers when the old syntax is used by the user. I'm a little bit hesitant to introduce that complex logic.

psigen: Oh, you raise a very good point.  The easiest way I can think to do this is to create a string sentinel value using a UUID that indicates that the file should not be generated and assign it here:
```
DO_NOT_RENDER = "CHIMERA_NO_RENDER_86c10773-c4e8-4b2f-95b3-137f8e4ea0da"
```
Then we can assign this value in the if-block and maybe check for equality in the `Render()` block?

This opens up a new possibility, which is that someone can set these keys to null if they want to suppress generation of either headers or cpp files for one of these templates.  I am not sure when this would be useful, maybe if they are using chimera in conjunction with another tool?

src/configuration.cpp
```diff
@@ -728,13 +791,49 @@ bool chimera::CompiledConfiguration::Render(
...
+    // from the configuration file's "template::file" entry.
+    if (bindingNode_)
+    {
+        chimera::util::extendWithYAMLNode(
+            full_context, bindingNode_["file"], false,
+            std::bind(&chimera::CompiledConfiguration::Lookup, this,
+                      std::placeholders::_1));
+    }
+
+    if (not Render(mangled_name, header_view, "h", context, full_context))
```
psigen, line 815: May we use the `!` here?  I don't know of any new codebases that use `not`, is this becoming popular somewhere?
```suggestion
    if (!Render(mangled_name, header_view, "h", context, full_context))
```

src/configuration.cpp
```diff
@@ -728,13 +791,49 @@ bool chimera::CompiledConfiguration::Render(
...
+    binding_names_.push_back(mangled_name);
+    return true;
+}
 
+bool chimera::CompiledConfiguration::Render(
+    const std::string &mangled_name, const std::string &view,
+    const std::string &extension,
+    const std::shared_ptr<::mstch::object> &context,
+    const ::mstch::map &full_context)
+{
```
psigen, line 831: At the beginning of this function, we could check if `view` matches a `NO_RENDER` sentinel value and return immediately.

src/configuration.cpp
jslee02: I'm using `not` in other codebase and started using it everywhere habitually. :sweat_smile: Let me replace all the use of `not` in chimera to `!`.

src/configuration.cpp
```diff
@@ -451,20 +453,80 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+            {
+                bindingDefinition_.class_cpp = Lookup(node);
+            }
+            else
+            {
+                if (const auto header_node = lookupYAMLNode(node, "header"))
+                    bindingDefinition_.class_h = Lookup(header_node);
+                if (const auto source_node = lookupYAMLNode(node, "source"))
+                    bindingDefinition_.class_cpp = Lookup(source_node);
+            }
```
psigen, line 468: With the sentinel value this would look more like:
```suggestion
            if (node.IsNull())  // This needs to be different than *undefined*, which means unspecified
            {
                bindingDefinition_.class_h = chimera::util::FLAG_NO_RENDER;
                bindingDefinition_.class_cpp = chimera::util::FLAG_NO_RENDER;
            }
            if (node.IsScalar())
            {
                bindingDefinition_.class_h = chimera::util::FLAG_NO_RENDER;
                bindingDefinition_.class_cpp = Lookup(node);
            }
            else
            {
                if (const auto header_node = lookupYAMLNode(node, "header"))
                    bindingDefinition_.class_h = Lookup(header_node);
                if (const auto source_node = lookupYAMLNode(node, "source"))
                    bindingDefinition_.class_cpp = Lookup(source_node);
            }
```

src/configuration.cpp
jslee02: Using a sentinel value sounds good to me, but this might still have the issue with the old syntax though because in the old syntax we don't have a chance to use the sentinel value.



> This opens up a new possibility, which is that someone can also set the template to `null` if they want to suppress generation of both headers and cpp files for one of these templates.

Not sure if I get what you mean. What's the difference between using a sentinel value and setting the template tag with `null`? Do you mean we provide both ways?

jslee02: Never mind, I got the answers from your [code snippet](https://github.com/personalrobotics/chimera/pull/289#discussion_r408625960).

src/configuration.cpp
jslee02: Done https://github.com/personalrobotics/chimera/pull/289/commits/3fb77d63aa41ac4da2e2680fb509ed101dc6a03a

src/configuration.cpp
jslee02: Done https://github.com/personalrobotics/chimera/pull/289/commits/3fb77d63aa41ac4da2e2680fb509ed101dc6a03a

include/chimera/util.h
```diff
@@ -20,6 +20,9 @@ namespace chimera
 namespace util
 {
 
+constexpr char FLAG_NO_RENDER[]
+    = "CHIMERA_NO_RENDER_86c10773-c4e8-4b2f-95b3-137f8e4ea0da";
```
jslee02, line 24: Not sure if this is what you meant.

psigen: I think so, there may be better ways to do this in recent C++ standards with `std::string` or `std::string_view`:
https://stackoverflow.com/a/43335753/2044807

jslee02: Thanks for the confirmation. Yeah, I was looking at the same post!

PR #292, opened 2020-04-13
Update clang type to string conversion helper function

jslee02:


src/util.cpp
```diff
@@ -968,10 +968,13 @@ std::string toString(QualType qual_type)
...
 std::string toString(const Type *type)
 {
+    // Reference:
+    // https://clang.llvm.org/doxygen/classclang_1_1ExtQualsTypeCommonBase.html
+
     if (dyn_cast<DecayedType>(type))
         return "DecayedType : AdjustedType";
     else if (dyn_cast<AdjustedType>(type))
-        return "ArrayType";
+        return "AdjustedType";
```
psigen, line 977: Good catch :+1:

src/util.cpp
```diff
@@ -1002,12 +1005,16 @@ std::string toString(const Type *type)
     else if (dyn_cast<DeducedTemplateSpecializationType>(type))
         return "DeducedTemplateSpecializationType : DecudedType";
     else if (dyn_cast<DeducedType>(type))
-        return "DeducedDecudedType";
+        return "DeducedType";
```
psigen, line 1008: Another good catch.

PR #295, opened 2020-04-14
Suppress #148: Skip generating binding for lambda function

jslee02:
This PR suppresses #148 by skipping generating class binding that is being generated for lambda function defined in a header. We can safely skip the binding because the function contains the lambda function (i.e., `foo()` in the test) is still generated.


jslee02:
@psigen Could you review this PR at your convenience? This is a critical bug for a project that contains lambda functions in template classes.

PR #298, opened 2020-04-14
Ignore unspecialized template variables

### LINKED ISSUE #297: Generating template variable
```cpp
#include <type_traits>
namespace chimera_test {
template <typename T>
constexpr bool template_var = std::true_type::value;  // [1]
constexpr bool specialized_var = template_var<int>;   // [2]
}
```
Generated binding for this source:
```cpp
m.attr("template_var") = chimera_test::template_var;       // [3] invalid binding
m.attr("specialized_var") = chimera_test::specialized_var  // [4] valid binding
m.attr("template_var") = chimera_test::template_var;       // [5] invalid binding
```

[3] is generated when `template_var` in [1] is visited, but the binding shouldn't be generated because the template variable is not specialized.

[4] and [5] are generated when `specialized_var` in [2] is visited. [4] is okay because `specialized_var` is just a global variable being assigned with a specialized template variable (i.e., `template_var<int>`). However, [5] should be either not generated or be `m.attr("template_var_int") = chimera_test::template_var<int>`, which I'm not sure which one would be the right way. If we need to generate [5], note that the variable name should be mangled name (e.g., `template_var_int`) unless specified by config (also, we don't have a way to specify properties for variables).

#298 changes Chimera to skip generating [3] and [5].

jslee02:
The variable template is a C++14 feature. We need to (1) check which LLVM version added function `decl->getDescribedVarTemplate()` and (2) enable the test only when Chimera is built with C++14 supported compiler.

Fixes #297

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -1131,9 +1131,11 @@ std::string toString(const Decl *decl)
     else if (dyn_cast<TypeDecl>(decl))
         return "TypeDecl : NamedDecl";
     // ...
+#if LLVM_VERSION_AT_LEAST(6,0,0)
     else if (dyn_cast<DecompositionDecl>(decl))
         return "DecompositionDecl : VarDecl : DeclaratorDecl : ValueDecl : "
                "NamedDecl";
+#endif
     else if (dyn_cast<ImplicitParamDecl>(decl))
         return "ImplicitParamDecl : VarDecl : DeclaratorDecl : ValueDecl : "
                "NamedDecl";
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -968,6 +968,10 @@ std::string toString(QualType qual_type)
 
 std::string toString(const Type *type)
 {
+    // Reference:
+    // https://clang.llvm.org/doxygen/classclang_1_1ExtQualsTypeCommonBase.html
+
+    // TODO: There are more types that are not handled.
     if (dyn_cast<DecayedType>(type))
         return "DecayedType : AdjustedType";
     else if (dyn_cast<AdjustedType>(type))
@@ -1080,15 +1084,15 @@ std::string toString(const Type *type)
 
 std::string toString(const Decl *decl)
 {
-    // TODO: Incomplete
+    // Reference: https://clang.llvm.org/doxygen/classclang_1_1Decl.html
+
+    // TODO: There are more types that are not handled.
     if (dyn_cast<AccessSpecDecl>(decl))
         return "AccessSpecDecl";
     else if (dyn_cast<BlockDecl>(decl))
         return "BlockDecl";
-    // ...
     else if (dyn_cast<LabelDecl>(decl))
         return "LabelDecl : NamedDecl";
-        // ...
 #if LLVM_VERSION_AT_LEAST(3, 9, 0)
     else if (dyn_cast<BuiltinTemplateDecl>(decl))
         return "BuiltinTemplateDecl : TemplateDecl : NamedDecl";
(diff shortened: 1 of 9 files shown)


jslee02:
This patch still doesn't work for this case:
```cpp
template <typename T>
constexpr bool template_var = std::true_type::value;
constexpr bool specialized_var = template_var<int>;
```
generating following undesired binding:
```cpp
attr.attr("template_var") = chimera_test::template_var;        // [1] fails to build
attr.attr("specialized_var") = chimera_test::specialized_var;  // [2] works fine
```

If `template_var<int>` is changed to `true` (or `false`), then [1] is not generated, which is the expected behavior.

jslee02:
Updated description of #297. This PR now simply ignores [3] and [4] in the example binding in #297.

src/util.cpp
```diff
@@ -1131,6 +1131,27 @@ std::string toString(const Decl *decl)
...
+    else if (dyn_cast<ParmVarDecl>(decl))
+        return "ParmVarDecl : VarDecl : DeclaratorDecl : ValueDecl : NamedDecl";
+    else if (dyn_cast<VarTemplateSpecializationDecl>(decl))
+        return "VarTemplateSpecializationDecl : VarDecl : DeclaratorDecl : "
+               "ValueDecl : NamedDecl";
+    else if (dyn_cast<VarDecl>(decl))
+        return "VarDecl : DeclaratorDecl : ValueDecl : NamedDecl";
+    else if (dyn_cast<DeclaratorDecl>(decl))
+        return "DeclaratorDecl : ValueDecl : NamedDecl";
+    // ...
```
psigen, line 1151: What does this ... mean?

src/visitor.cpp
```diff
@@ -237,6 +237,13 @@ bool chimera::Visitor::GenerateGlobalVar(clang::VarDecl *decl)
         return false;
     else if (!decl->isThisDeclarationADefinition())
         return false;
+    // Ignore unspecialized template variables.
+    else if (decl->getDescribedVarTemplate())
+        return false;
+    // TODO(#297): Ignore specialized template variables because the current
+    // variable mastch and template not able to generate valid binding.
```
psigen, line 244: What is `mastch`?

src/util.cpp
jslee02: That denotes that there are more types between previous and next types according to the LLVM [manual](https://clang.llvm.org/doxygen/classclang_1_1Decl.html). This is because I'm lazy to add all the types. :sweat:

src/visitor.cpp
jslee02: I meant `mstch` or `mustache` trying to refer to the `::match::object`. What would be a better notion?

src/util.cpp
psigen: Ok, I think just put a small comment saying there are more types that are not handled.

src/visitor.cpp
psigen: Oh ok, then `mstch` is fine, it is just a typo.

src/visitor.cpp
```diff
@@ -244,6 +244,13 @@ bool chimera::Visitor::GenerateGlobalVar(clang::VarDecl *decl)
...
```
psigen, line 251:
```suggestion
    // variable mstch template not able to generate valid binding.
```

src/visitor.cpp
```diff
@@ -244,6 +244,13 @@ bool chimera::Visitor::GenerateGlobalVar(clang::VarDecl *decl)
         return false;
     else if (!decl->isThisDeclarationADefinition())
         return false;
+    // Ignore unspecialized template variables.
+    else if (decl->getDescribedVarTemplate())
+        return false;
+    // TODO(#297): Ignore specialized template variables because the current
+    // variable mastch template is not able to generate valid binding.
```
psigen, line 251:
```suggestion
    // variable mstch template is not able to generate valid binding.
```

psigen:
> This patch still doesn't work for this case:

Should we still merge as-is, or try to fix this case?

jslee02:
> Should we still merge as-is, or try to fix this case?

It's fixed by commits after the comment. https://github.com/personalrobotics/chimera/pull/298#issuecomment-613534822 and https://github.com/personalrobotics/chimera/issues/297 are the most up-to-date status of the issue. So we can merge as it is.

PR #304, opened 2020-04-15
Add an option to get list of generated bindings

jslee02:
Currently, no way to easily get the list of generated binding from CMake function `add_chimera_binding(...)`. [We had to collect the list of files by reading the binding directory](https://github.com/dartsim/dart/blob/e3823a2f6d5b5b482838d85bf3a0cae14cebce67/python/dartpy/CMakeLists.txt#L116-L119), which is not convenient.

This PR adds an optional output argument to `add_chimera_binding(...)` that will be assigned the list of generated bindings.


psigen:
Is this a list of targets in cmake names, or a list of source files?  I'm not sure which output this is...

jslee02:
It's a list of source files. I didn't use the term "source" because it could be confused with the source for generating binding (source) files. Feel free to suggest better option name and description :)

jslee02:

cmake/chimeraFunctions.cmake
```diff
@@ -24,7 +25,7 @@ function(add_chimera_binding)
     # Unparsed arguments can be found in variable ARG_UNPARSED_ARGUMENTS.
     set(prefix binding)
     set(options DEBUG EXCLUDE_FROM_ALL)
-    set(oneValueArgs TARGET MODULE CONFIGURATION DESTINATION BINDING)
+    set(oneValueArgs TARGET MODULE CONFIGURATION DESTINATION BINDING GENERATED_BINDING_SOURCES)
```
psigen, line 28: Please rename this to `GENERATED_SOURCES_VAR` to help distinguish it as a string variable name rather than a value.

Also, does this differ in a meaningful way from the result of  ```get_target_property(GENERATED_SOURCES my_chimera_target SOURCES)```

jslee02: > Please rename this to `GENERATED_SOURCES_VAR` to help distinguish it as a string variable name rather than a value.

Sounds good to me



> Also, does this differ in a meaningful way from the result of `get_target_property(GENERATED_SOURCES my_chimera_target SOURCES)`

I believe it would return all the sources including the sources specified as `EXTRA_SOURCES` for this function. This variable is to replace [this hassle](https://github.com/personalrobotics/dartpy/blob/3535357b354b48919d9912c24aff37eff8d6e711/CMakeLists.txt#L144-L147) in the user side.

psigen: But there, we would be able to use the `EXTRA_SOURCES`, since we need them anyway to compile, right?

jslee02: That's correct.

I think this variable still useful as it's an easier way than using the extra function. I'm okay to remove this variable and document about how to extract the source files.

The bigger motivation behind this variable is to simplify the chimera setting on the user side (#317). If we extend `add_chimera_binding()` to create the custom target just for generating bindings, then this variable wouldn't be much useful.

jslee02: > The bigger motivation behind this variable is to simplify the chimera setting on the user side (#317). If we extend `add_chimera_binding()` to create the custom target just for generating bindings, then this variable wouldn't be much useful.

If this sounds good to you, then I think we can revert this PR (or when #317 is resolved).

Issue #302, opened 2020-04-15
Format generated bindings using ClangFormat if available


jslee02:
Generated bindings are generally not well-formatted. We could format the code using ClangFormat if detected.

psigen:
We should not have to do this if we can simply appropriately adjust the templates.

jslee02:
There could be some limitations on that approach such as the column limit.

I understand that code formatting can be done on the user side, and it's one of the motivations of #304 to get the list of generated files for formatting.

jslee02:
This feature would be very useful for the case that chimera is not currently not able to generate for the entire C++ codebase for some reason but still used to generate partial bindings. In this case, the user might need to modify the generated bindings manually. It would be less convenient to manage unformatted code (imagine chimera generates very long length of method binding with long mangled names, options, etc).

PR #309, opened 2020-04-17
Update pybind11 class template to allow overriding virtual functions of parent classes

jslee02:
This PR allows C++ virtual functions declared in the ancestor classes to be overridden in Python. Without this change, we have to override all the ancestor virtual functions in the "trampoline" class, which is currently not supported by `chimera`.

However, this method also has a limitation that this won't work for multiple inheritances due to the diamond inheritance issue. The diamond inheritance issue won't happen if the class template provide a field for the list of all the virtual functions defined in all the ancestor classes.

See the [pybind11 doc](https://pybind11.readthedocs.io/en/stable/advanced/classes.html#combining-virtual-functions-and-inheritance) for the details.


jslee02:
Yes, I was trying to fix indentations, but the change wasn't perfect. I'm more into fixing the format issue by https://github.com/personalrobotics/chimera/issues/302.

Issue #310, opened 2020-04-18
pybind11 doesn't allow same name nested class defined in different scope


jslee02:
Given following code, pybind11 doesn't allow to register `Base::Option` and `Derived::Option` complaining "Option" is already defined. Boost.Python works fine.
```cpp
class Base
{
public:
    Base() = default;
    class Option
    {
    public:
        Option() = default;
    };
};

class Derived : public Base
{
public:
    Derived() = default;
    class Option : public Base::Option
    {
    public:
        Option() = default;
    };
};
```

jslee02:
We can avoid this by specifying the name in the configuration YAML, but it would be great to provide options to automatically handle this issue.

Possible options:

- O1: Prepend the parent class name to the nested class
  ```cpp
  base_option = Base.BaseOption()
  derived_option = Derived.DerivedOption()
  ```
- O2: Same as O1 but move the nested class to the parent scope
  ```cpp
  base_option = BaseOption()
  derived_option = DerivedOption()
  ```
- O3: Same as O2 but put underscore between class names
  ```cpp
  base_option = Base_Option()
  derived_option = Derived_Option()
  ```

Additionally, we may want to detect the name confliction to catch this error before generating bindings and either not generating the binding or renaming it to a unique name like appending characters (e.g., multiple underscores or numbers).

jslee02:
@psigen Would you have some suggestions on this? This is another name conflict issue that might need the scope-changing method or/and name mangling method.

jslee02:
Just in case this is pybind11 issue: https://github.com/pybind/pybind11/issues/2187

PR #311, opened 2020-04-18
Add regression test for #310

jslee02:


jslee02:

PR #312, opened 2020-04-19
Add typedef support

jslee02:
Typedef and type alias are currently ignored. This PR adds typedef support inspired by [this SO post](https://stackoverflow.com/a/53685393/3122234).

Disclaimer: This PR only works for typedef for the case that satisfying that (1) the original type is not templated type (since chimera doesn't generate bindings for templated classes) and (2) the original type is the type whose the binding is being generated.

Both limitations will be resolved by future PRs.


src/util.cpp
```diff
@@ -1120,6 +1126,9 @@ std::string toString(const Decl *decl)
 {
     // Reference: https://clang.llvm.org/doxygen/classclang_1_1Decl.html
 
+    if (decl == nullptr)
+        return "Error: nullptr";
```
psigen, line 1130: Is returning this malformed string the best thing to do?  It seems like we should at least put some boundary around it like `[nullptr]` because `:` is used as a valid delimiter in the other strings.

test/examples/07_typedef/typedef.cpp
```diff
@@ -0,0 +1,8 @@
+#include "typedef.h"
+
+namespace chimera_test
+{
+
+//==============================================================================
+
+} // namespace chimera_test
```
psigen, line 8: Do we even need this file?  Everything should work without it right?

jslee02: We don't. It seems I add source files habitually even it's empty. Let me remove this.

src/util.cpp
jslee02: Sounds good to me. I didn't pay attention to the string pattern as I use this just for debugging.

jslee02:
> Is this PR ready for review?

Yes!

src/util.cpp
psigen: Should this throw some sort of message to `std::cerr` as per other warnings?

jslee02: Done: https://github.com/personalrobotics/chimera/pull/312/commits/ee496181137b95614226df7751d2ec26e741fa27

src/util.cpp
```diff
@@ -1120,6 +1126,14 @@ std::string toString(const Decl *decl)
 {
     // Reference: https://clang.llvm.org/doxygen/classclang_1_1Decl.html
 
+    if (decl == nullptr)
+    {
+        std::cerr << "Warning: nullptr is not allowed to get the type name of "
+                     "clang::Decl*"
+                  << std::endl;
+        return "Error: [nullptr]";
```
psigen, line 1134: This is being used _in_ other template strings, we really want to put a delimiter around this so we can see this clearly in the templates where it would show up.

```suggestion
        return "[nullptr]";
```

```suggestion
        return "std::nullptr_t";
```
I don't see a point in putting "error" in the string itself, since we are already printing the warning.  Putting a verbose error string into the rendered binding is just going to cause even more trouble.

Is there a compiler preprocessor directive or static_assert we could return that would do something meaningful when compiled?

jslee02: > This is being used in other template strings

Hm, I'm not able to find a place this function is used in the core implementation. Could you point me where it is?

psigen: Hmm, I thought it was for the templates, but maybe it is only for type lookups or error messages.

psigen: Actually, now that you mention it, what is the case where we hit this?

Should it just be a `std::runtime_error`?  It seems like we would only get `nullptr` in an actual bug situation.

jslee02: We have a dedicated function to get the type name, and these functions are created just for debugging for the case that the dedicated function doesn't work for some reason. I documented this in the header. Should we move these functions to a different namespace or a nested namespace (e.g., `chimera::util::debug`) to be more clear that this is only for debugging?

jslee02: > Actually, now that you mention it, what is the case where we hit this?

Well, this could hit when I pass `nullptr` in debugging? :thinking: 

> Should it just be a `std::runtime_error`? It seems like we would only get `nullptr` in an actual bug situation. We cannot get it spuriously due to bad input, right?

Not sure, if we really need to consider how to deal with `nullptr` because this function is only temporarily used during debugging, and we never commit a code that uses this function.

Issue #121, opened 2017-03-13
Invalid code generation for anonymous enums


jslee02:
For anonymous enums, Chimera generates invalid code such as `dart::common::ClassName::(anonymous)::EnumValue`. Instead, it would be nice to print a warning and skip the code generation for the anonymous enums.

jslee02:
This also happens for an anonymous namespace. The generated code looks like `dart::math::(anonymous namespace)::MyClass::foo()`.

jslee02:
Anonymous enum is now ignored by #121 as a workaround. A better solution would be defining the elements of an anonymous enum in its parent scope with valid binding code.

PR #313, opened 2020-04-20
Ignore anonymous enum (#121)

jslee02:
For an anonymous enum, chimera generates an invalid mangled name (i.e., file name contains special character like `$`) and invalid binding code (i.e., type name of `(anonymous)`) to compile the binding.

This PR changes chimera to ignore anonymous enum as a workaround. A better solution would be defining the elements of an anonymous enum to its parent scope.


jslee02:

PR #318, opened 2020-04-26
Rename GENERATED_BINDING_SOURCES with GENERATED_SOURCES_VAR

jslee02:


jslee02:
Let me merge this PR since this is just a variable name renaming.

Let's continue discussing whether to keep this variable in #304.

PR #320, opened 2020-04-27
Use std::cerr for warnings

jslee02:
Otherwise, the warnings are regarded as the binding file name, which is an error.


jslee02:
Merging as this is a simple bug fix

PR #321, opened 2020-04-27
Skip static methods that overload instance methods

### LINKED ISSUE #262: Shouldn't generate both of static and non-static method with the same name for pybind11

jslee02:
This PR suppresses static methods that overload instance methods with the same name.

Fixes #262.


src/mstch.cpp
```diff
@@ -478,55 +479,67 @@ ::mstch::node CXXRecord::methods()
...
+
     ::mstch::array method_templates = boost::get<::mstch::array>(methods());
-    std::map<std::string, bool> is_static_method;
 
-    // Iterate through all methods searching for static ones.
+    // Add all static method name to this list.
     for (const ::mstch::node method_node : method_templates)
     {
-        // Resolve the static-ness and name of each function from method
+        // Resolve the static-ness  and name of each function from method
```
psigen, line 531: nit: whitespace
```suggestion
        // Resolve the static-ness and name of each function from method
```

src/mstch.cpp
```diff
@@ -478,55 +479,67 @@ ::mstch::node CXXRecord::methods()
...
 
 ::mstch::node CXXRecord::staticMethods()
 {
+    ::mstch::array static_method_list;
+
     ::mstch::array method_templates = boost::get<::mstch::array>(methods());
-    std::map<std::string, bool> is_static_method;
 
-    // Iterate through all methods searching for static ones.
+    // Add all static method name to this list.
```
psigen, line 528:
```suggestion
    // Add all static method names to this list.
```

src/mstch.cpp
```diff
@@ -478,55 +479,67 @@ ::mstch::node CXXRecord::methods()
...
+
+        method_templates.push_back(method);
+    }
+
     return method_templates;
 }
 
 ::mstch::node CXXRecord::staticMethods()
 {
+    ::mstch::array static_method_list;
```
psigen, line 524: Let's move this declaration to right above the `for (const ::mstch::node method_node : method_templates)` where it is used.

src/mstch.cpp
```diff
@@ -478,55 +479,67 @@ ::mstch::node CXXRecord::methods()
...
+    // Collect all the non-static method names
+    std::unordered_set<std::string> non_static_method_names;
+    for (const std::shared_ptr<Method> &method : method_vector)
+    {
+        const bool is_static
+            = (::mstch::render("{{#is_static}}true{{/is_static}}", method)
+               == "true");
+
+        if (!is_static)
+            non_static_method_names.insert(method->nameAsString());
```
psigen, line 491: I think this construction would be simpler, because it will make the next block simpler as well.

```suggestion
        const bool is_static = boost::get<bool>(method->at("is_static"));
        if (is_static)
            continue;

        non_static_method_names.insert(method->nameAsString());
```

src/mstch.cpp
```diff
@@ -478,55 +479,67 @@ ::mstch::node CXXRecord::methods()
...
+                std::cerr
+                    << "Warning: Skipping static method '" << name
+                    << "' because there is instance method(s) with the same "
+                    << "name. Consider renaming the conflicting static method "
+                    << "in the configuration YAML.\n";
+                continue;
+            }
+        }
+
+        method_templates.push_back(method);
```
psigen, line 516:
```suggestion
        const bool is_static = boost::get<bool>(method->at("is_static"));
        if (!is_static)
            continue;

        const std::string name = method->nameAsString();
        if (non_static_method_names.find(name) != non_static_method_names.end())
        {
            std::cerr
                << "Warning: Skipping static method '" << name
                << "' because there is instance method(s) with the same "
                << "name. Consider renaming the conflicting static method "
                << "in the configuration YAML.\n";
            continue;
        }
 
        method_templates.push_back(method);
```

src/mstch.cpp
jslee02: Done

src/mstch.cpp
```diff
@@ -478,18 +479,48 @@ ::mstch::node CXXRecord::methods()
...
-        method_templates.push_back(method_template);
+    // Collect all the non-static method names
+    std::unordered_set<std::string> non_static_method_names;
+    for (const std::shared_ptr<Method> &method : method_vector)
+    {
+        const bool is_static = boost::get<bool>(method->at("is_static"));
+        if (is_static)
+            continue;
+
+        non_static_method_names.insert(method->nameAsString());
```
psigen, line 490: Oops, wait, do we need this here too?
```suggestion
        non_static_method_names.insert(method->nameAsString());
        method_templates.push_back(method);
```

src/mstch.cpp
```diff
@@ -478,18 +479,48 @@ ::mstch::node CXXRecord::methods()
     if (method_vector.size() > 0)
         method_vector.back()->setLast(true);
 
-    // Copy each template into the mstch template array.
-    for (auto method_template : method_vector)
-        method_templates.push_back(method_template);
+    // Collect all the non-static method names
```
psigen, line 482:
```suggestion
    // Copy each non-static method into the mstch template.
```

src/mstch.cpp
```diff
@@ -478,18 +479,48 @@ ::mstch::node CXXRecord::methods()
...
+    {
+        const bool is_static = boost::get<bool>(method->at("is_static"));
+        if (is_static)
+            continue;
+
+        non_static_method_names.insert(method->nameAsString());
+    }
+
+    // Copy each template into the mstch template array except static methods
+    // that have the same name with instance methods.
```
psigen, line 494:
```suggestion
    // Copy each static method into the mstch template array, except those
    // that have the same name as an instance method.
```

psigen:
Hmm, I was thinking about this more and I think we might be making this too complicated.

Instead of suppressing the methods at all, why don't we do the following:
* `methodsInternal()`: *(private method)* returns all methods -- what `method()` does now, with no suppression -- it can have duplicates.
* `methods()`: returns all instance methods (filter by `!is_static`)
* `staticMethods()`: returns all static methods (filter by `is_static`)

Then we can simply change the templates slightly, because I don't think there is a use-case where someone wants both static and instance methods at the same time!

psigen:
The reason I think this is better is because if someone wants just the static methods, they probably don't want some of them to be suppressed...

jslee02:
:+1: It makes sense to me.

One concern is that the current `staticMethods()`, which is being used in the Boost.Python template, returns the names of static methods (not the template). I think it's safe to overwrite the function to return the static method templates because we can get the names anyway (just need to change the template).

psigen:
Yeah I agree, this is my intent.  It is easier to have several view methods that all return templates, I think it should work like `fields()` and `staticFields()`

src/mstch.cpp
```diff
@@ -603,6 +569,77 @@ ::mstch::node CXXRecord::staticFields()
     return static_field_templates;
 }
 
+::mstch::node CXXRecord::methodsInternal() const
```
jslee02, line 572: Exactly the same function (but the name) as the previous `methods()`.

include/chimera/mstch.h
```diff
@@ -218,14 +218,31 @@ class CXXRecord : public ClangWrapper<clang::CXXRecordDecl>
...
+     */
     ::mstch::node methods();
+    /**
+     * Returns all the static methods
+     */
     ::mstch::node staticMethods();
+    /**
+     * Returns all the instance methods and the static methods except the
+     * static methods whose name conflict with instance methods.
+     */
```
psigen, line 232: :tada:

include/chimera/mstch.h
```diff
@@ -218,14 +218,31 @@ class CXXRecord : public ClangWrapper<clang::CXXRecordDecl>
...
 
     ::mstch::node fields();
     ::mstch::node staticFields();
 
 protected:
     const std::set<const clang::CXXRecordDecl *> *available_decls_;
+
+private:
+    /**
+     * Returns all the methods.
```
psigen, line 243:
```suggestion
     * Returns all the methods.
     * This may include instance methods and static methods with conflicting names.
```

src/mstch.cpp
```diff
@@ -415,119 +419,81 @@ ::mstch::node CXXRecord::constructors()
...
-        }
+        // Add non-static method template to the visible method template list
+        visible_methods.push_back(method_node);
     }
 
-    // Add all static method names to this list.
-    ::mstch::array static_method_list;
-    for (auto const &entry : is_static_method)
+    // Add all non-static method templates to this list.
+    ::mstch::array non_static_methods;
```
psigen, line 482: This is unused now, right?

jslee02: Yes, removed

src/mstch.cpp
```diff
@@ -415,119 +419,75 @@ ::mstch::node CXXRecord::constructors()
...
-            is_static_method[name] = is_static;
-        }
+        visible_methods.push_back(method_node);
     }
 
-    // Add all static method names to this list.
-    ::mstch::array static_method_list;
-    for (auto const &entry : is_static_method)
+    // Copy each static method into the mstch template if the static method name
+    // doesn't conflict with the non-static method names
```
psigen, line 477:
```suggestion
    // doesn't conflict with the non-static method names.
```

jslee02:
> Awesome! Thank you for iterating with me on this, it looks very nice now.

I also like the current version much better! :+1:

test/examples/regression/issue262_static_method/issue262_static_method.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def _test_issue262(self, binding):
+        # Should be skipped as this static method overloads instance method
+        if binding is py11:
+            with self.assertRaises(TypeError):
+                binding.Integer.add(1, 2)
+        elif binding is boost:
+            # For some reason, Boost.Python allows instance and static method overloading
+            # For the details see: https://github.com/personalrobotics/chimera/issues/262
+            self.assertEqual(binding.Integer.add(1, 2), 3)
```
jslee02, line 17: @psigen Do you think we should disable instance and static method overloading for both bindings? For some reason, Boost.Python allows this, which I don't know why because it shouldn't be supported it's not valid Python syntax I believe.

jslee02: Never mind. Boost.Python also doesn't allow this case as well. Updated the templates and tests.

Issue #57, opened 2015-12-04
Wrap overloaded operators


mkoval:
These are currently suppressed.

jslee02:
Closing as operator overloading is supported by #322

PR #322, opened 2020-04-27
Add basic operator overloading support

jslee02:
This is a WIP PR to get early feedback.

pybind11 example: https://pybind11.readthedocs.io/en/stable/advanced/classes.html#operator-overloading
Boost.Python example: https://www.boost.org/doc/libs/1_66_0/libs/python/doc/html/tutorial/tutorial/exposing.html#tutorial.exposing.class_operators_special_function.python_operators

We could use simpler API using `{boost::python|pybind11}::self`, but saving it for future PRs if needed as it would require adding some code complexity to chimera to support it.

Related issue: #57


src/mstch.cpp
```diff
@@ -924,6 +932,21 @@ ::mstch::node Function::usesDefaults()
...
+std::string Function::nameAsString()
+{
+    if (decl_->isOverloadedOperator())
+    {
+        std::string operatorName;
+        if (chimera::util::getPythonOperatorSpelling(
+                decl_->getOverloadedOperator(), operatorName))
+        {
+            return operatorName;
+        }
```
psigen, line 944:
```suggestion
        return chimera::util::getPythonOperatorSpelling(decl_->getOverloadedOperator());
```

include/chimera/util.h
```diff
@@ -311,6 +311,14 @@ std::pair<unsigned, unsigned> getFunctionArgumentRange(
  */
 bool hasNonPublicParam(const clang::CXXMethodDecl *decl);
 
+/**
+ * Converts clang operator type to the operator name in Python.
+ *
+ * Return true on success.
+ */
+bool getPythonOperatorSpelling(clang::OverloadedOperatorKind kind,
+                               std::string &str);
```
psigen, line 320: I propose the following type that always returns some string.  Since we already have a check for which kinds of operators we support, there is no point in doing the check twice.

We could possibly raise a runtime error in the `default` case since this would be an internal logic error if the two did not match anyway.

```suggestion
std::string getPythonOperatorName(clang::OverloadedOperatorKind kind);
```

src/util.cpp
```diff
@@ -929,6 +929,14 @@ bool hasNonPublicParam(const CXXMethodDecl *decl)
...
             CXXRecordDecl *cxx_record_decl
                 = cast<CXXRecordDecl>(record_type->getDecl());
+
+            // Allow the parent class, in which the method is defined. In this
+            // case, the access specifier is As_none.
+            if (decl->getParent() == cxx_record_decl)
+            {
+                continue;
+            }
+
```
psigen, line 939: Hmm, I am not sure why this needed to be modified?

src/util.cpp
```diff
@@ -938,6 +946,22 @@ bool hasNonPublicParam(const CXXMethodDecl *decl)
...
+bool getPythonOperatorSpelling(OverloadedOperatorKind kind, std::string &str)
+{
+    switch (kind)
+    {
+        case OverloadedOperatorKind::OO_Plus:
+            str = "__add__";
+            return true;
+        case OverloadedOperatorKind::OO_Star:
+            str = "__mul__";
+            return true;
```
psigen, line 958: I wonder if there is any way we can resolve this to a lookup in the binding configuration.  I think that would be best so that different bindings could map these in different ways.  But I am not sure what the best approach is.

include/chimera/util.h
jslee02: > We could possibly raise a runtime error in the `default` case since this would be an internal logic error if the two did not match anyway.

This implies that we raise runtime error for unsupported or implemented operator types. Is this the intention? Not sure if I get what you mean by "the two did not match anyway".

psigen: We already have this `if` block which is checking which operators we pass to this function anyway:
https://github.com/personalrobotics/chimera/pull/322/files#diff-3f8c8328f0a509fbeb0a6d422bf8a85bR443

So I don't see the point in _also_ checking the same thing inside the `getPythonOperatorName()` function, since we can just make sure we implement anything that makes it past this check anyway.  Then if we hit the default case, we must have some sort of bug where things are getting past the check that should not be.

src/util.cpp
jslee02: The access type is `AS_none` for a parameter type in a method whose parent class is the parameter type itself. Without this block, we ignore all the methods that take a type of their parent class type, which I don't think what we want. For example,
```cpp
class Vector3d {
public:
  Vector3d operator+(const Vector3d& other);  // This method is ignored because `cxx_record_decl->getAccess() == AS_none` in this case
};
```

src/mstch.cpp
```diff
@@ -435,7 +435,14 @@ ::mstch::node CXXRecord::methods()
...
             continue;
         if (method_decl->isOverloadedOperator())
-            continue;
+        {
+            const OverloadedOperatorKind kind
+                = method_decl->getOverloadedOperator();
+            // TODO: Support more operators
+            if (kind != OverloadedOperatorKind::OO_Plus
+                && kind != OverloadedOperatorKind::OO_Star)
+                continue;
```
psigen, line 444: While it is a bit more verbose, I think a `switch` might be clearer as we have more operators.
```suggestion
            switch (kind)
            {
                case OverloadedOperatorKind::OO_Plus:
                case OverloadedOperatorKind::OO_Star:
                     break;  // Allow these kinds.
                default:
                    continue;  // Suppress any other kinds.
            }
```

src/util.cpp
psigen: Ah ok.  C++ is full of so many quirks 😓 

The documentation helpfully says:
https://clang.llvm.org/doxygen/Specifiers_8h_source.html#l00117
> A C++ access specifier (public, private, protected), plus the
> special value "none" **which means different things in different contexts.**

🤦

src/util.cpp
jslee02: I basically like this idea as it would make chimera more flexible for broader language binding support as you pointed out. Just don't have a good idea how to implement. Especially adding a map to the template wouldn't be straightforward since mustache is a logic-less template.

Maybe using a new configuration YAML block or file for language specifics?

include/chimera/util.h
jslee02: Sounds fair to me!

src/util.cpp
jslee02: Yeah, it should have been named with something like `AS_different_things_in_different_contexts` so that the user won't underestimate it what it can mean. :joy:

jslee02:
> The only thing I am trying to figure out is how to make the naming convention of e.g. `__add__` be specified in the binding template rather than in the C++ code.

Maybe we accept this approach for now and create an issue for the case that we want to support more than Python binding?

jslee02:
> That is because if I wanted to implement e.g. a nodejs binding language, this would not use the same names

It'd be awesome to see chimera works for more than one language support!

include/chimera/util.h
```diff
@@ -315,6 +315,11 @@ std::pair<unsigned, unsigned> getFunctionArgumentRange(
  */
 bool hasNonPublicParam(const clang::CXXMethodDecl *decl);
 
+/**
+ * Converts clang operator type to the operator name in Python.
+ */
+std::string getPythonOperatorName(clang::OverloadedOperatorKind kind);
```
psigen, line 321: Let's just call this `getOperatorName()` under the assumption it will not be python-specific later.

PR #324, opened 2020-04-28
Add module docstring support

### LINKED ISSUE #285: Provide a way to set module docstring
Chimera extract comments from binding objects (e.g., class and function), but there is no way to set the docstring for the top level module.

jslee02:
Fixes #285


src/configuration.cpp
```diff
@@ -231,6 +231,22 @@ chimera::CompiledConfiguration::CompiledConfiguration(
...
+            // Check that 'module' node in configuration YAML is a map.
+            if (!docNode.IsScalar())
+            {
+                throw std::runtime_error(
+                    "'module.doc' in configuration YAML must be a scalar.");
+            }
+
+            module_doc_ = docNode.as<std::string>();
+        }
+
```
psigen, line 249: Wait.. doesn't this work without a lookup?  I thought this value would just get passed through with no C++ changes because of the dictionary merge?

jslee02: It should work. The feature keeps slipping my mind. :(

jslee02: It seems not working without C++: https://dev.azure.com/personalrobotics/chimera/_build/results?buildId=677&view=logs&j=64b073a0-3b10-5573-3eb8-84d1090a7daf&t=d1aadda8-ef7c-5bd9-cfe7-1872f1f8342b&l=1717

psigen: Hmm, there is probably something missing then.  We should look into how this is different than the other instances, maybe some code is missing to merge the values.

jslee02: It seems this is because we extend the YAML node from `template.file`?
https://github.com/personalrobotics/chimera/blob/c43e0cf300b2151d638a56b4bffb7e3c7dd34ccb/src/configuration.cpp#L742-L748

The docstring for the module is added to `module.doc`, which is out of `template` scope in the YAML configuration.

psigen: Maybe it is expecting to use `main` instead?  Or we need an additional block here?
https://github.com/personalrobotics/chimera/blob/master/src/configuration.cpp

jslee02: > Maybe it is expecting to use `main` instead? Or we need an additional block here?

I see three options:

(1) Add binding specific code to `template.main.precontent`
```yaml
template:
  main:
    precontent: |
      m.doc() = "<module_description>"  # or
      ::boost::python::scope().attr("__doc__") = "<module_description>";
```
(2) Add new tag `template.main.doc` and update the module template to use `{{doc}}` instead of `{{module.doc}}`

(3) Add new tag `module.doc` (as in this PR) and add additional block here something like:
```cpp
    // Resolve customizable snippets that will be inserted into the file
    // from the configuration file's "module" entry.
    if (const auto moduleNode = lookupYAMLNode(configNode_, "module"))
    {
        chimera::util::extendWithYAMLNode(
            module_map, moduleNode, true,
            std::bind(&chimera::CompiledConfiguration::Lookup, this,
                      std::placeholders::_1));
    }

    // Create a top-level context that contains the extracted information
    // about the module.
    ::mstch::map full_context{{"module", module_map}};
```

I'd prefer (2) because (1) is a bit fragile and (3) is redundant as we already have `template.main`.

psigen: (2) seems fine to me, it makes sense since `main` is about the same scope as module anyway.  

Separate question: should we rename `main` to `module`?  Is it more consistent with the template naming?

jslee02: > (2) seems fine to me, it makes sense since `main` is about the same scope as module anyway.

:+1: 



> Separate question: should we rename `main` to `module`? Is it more consistent with the template naming?

It sounds good to me!

jslee02: https://github.com/personalrobotics/chimera/pull/325

PR #327, opened 2020-05-05
Add initial support of typedef for built-in types

jslee02:
The type alias for built-in types is currently ignored when it's valid syntax in Python. 

```cpp
using Double = double;  // ignored by chimera
```

```python
>>> Double = float
>>> a = Double()
>>> b = float()
>>> type(a)
<class 'float'>
>>> type(b)
<class 'float'>
>>> Double == float
True
```

This PR is in prototype stage for the implementation discussion. The approach is to use `__main__` module as:
```cpp
m.attr("Double") = ::pybind11::module::import("__main__").attr("__builtins__").attr("float");
```
The compiled binding works as expected as
```python
>>> import typedef_pybind11 as py11
>>> py11.Double
<class 'float'>
>>> py11.Double == float
True
```


src/visitor.cpp
```diff
@@ -335,11 +335,23 @@ bool chimera::Visitor::GenerateTypedefName(TypedefNameDecl *decl)
 #endif
     }
 
+    // Create mstch for typedef where the underlying type is a builtin type
```
psigen, line 338:
```suggestion
    // Create mstch for typedef where the underlying type is a builtin type.
```

src/visitor.cpp
```diff
@@ -335,11 +335,23 @@ bool chimera::Visitor::GenerateTypedefName(TypedefNameDecl *decl)
...
+        return config_.Render(context);
+    }
+
     // Get the declaration from the underlying type
     auto resolved_decl = chimera::util::resolveRecord(
         config_.GetCompilerInstance(), underlying_type_name);
 
-    // Skip if failed to get the declaration or the declaration is not of class
+    // Skip if failed to get the declaration or the declaration is not of
+    // class
```
psigen, line 354:
```suggestion
    // Skip if we could not get the declaration, or the declaration was not
    // of a class.
```

src/util.cpp
```diff
@@ -953,6 +953,20 @@ bool hasNonPublicParam(const CXXMethodDecl *decl)
     return false;
 }
 
+std::string getPythonBuiltinTypeSpelling(const std::string &type_name)
+{
+    if (type_name == "double" || type_name == "float")
+        return "float";
```
psigen, line 959: Should we just do all the builtin types?  Most are pretty simple. :shrug: 
https://docs.microsoft.com/en-us/cpp/cpp/fundamental-types-cpp?view=vs-2019

include/chimera/util.h
```diff
@@ -315,6 +315,8 @@ std::pair<unsigned, unsigned> getFunctionArgumentRange(
  */
 bool hasNonPublicParam(const clang::CXXMethodDecl *decl);
 
+std::string getPythonBuiltinTypeSpelling(const std::string &type_name);
```
psigen, line 318: We should make this match whatever function name we end up using for the operator overloading.  It might have been `Name` instead of `Spelling`.

src/util.cpp
jslee02: I think so. It won't be a 1to1 map though.

include/chimera/util.h
jslee02: Sounds good to me. Let me rename them both with `Name`.

include/chimera/util.h
```diff
@@ -315,6 +315,11 @@ std::pair<unsigned, unsigned> getFunctionArgumentRange(
  */
 bool hasNonPublicParam(const clang::CXXMethodDecl *decl);
 
+/**
+ * Converts a C++ built-in type name to the Python equivalent type name.
+ */
+std::string getPythonBuiltinTypeName(const std::string &type_name);
```
psigen, line 321: Similarly, let us simply remove "Python" from this, as it is really just an arbitrary mapping as far as Chimera is concerned.

```suggestion
std::string getBuiltinTypeName(const std::string &type_name);
```

include/chimera/mstch.h
```diff
@@ -390,11 +390,26 @@ class Typedef : public ClangWrapper<clang::TypedefNameDecl>
...
 
+class BuiltinTypedef : public ClangWrapper<clang::TypedefNameDecl>
+{
+public:
+    BuiltinTypedef(const ::chimera::CompiledConfiguration &config,
+                   const clang::TypedefNameDecl *decl,
+                   const clang::BuiltinType *builtin_type);
+
+    ::mstch::node isBuiltinType();
+    ::mstch::node underlyingBuiltinType();
```
psigen, line 407: Perhaps this can be `underlyingType()`, since we use `type()` as an accessor on other wrappers.

```suggestion
    ::mstch::node underlyingType();
```

Issue #328, opened 2020-05-05
Fails to resolve dependent name type


jslee02:
chimera fails to resolve [`clang::DependentNameType`](https://clang.llvm.org/doxygen/classclang_1_1DependentNameType.html) such as `typename S::Scalar` in the following code:
```cpp
template<typename S>
struct ScalarTrait {
  typedef typename S::Scalar type;
};
```
that throws an exception as
```console
chimera.util.resolveDeclaration:1:18: error: use of undeclared identifier 'S'
typedef typename S::Scalar chimera_placeholder_4;
                 ^
chimera.util.resolveDeclaration:1:21: error: expected a qualified name after 'typename'
typedef typename S::Scalar chimera_placeholder_4;
                    ^
chimera.util.resolveDeclaration:1:27: error: expected ';' after top level declarator
typedef typename S::Scalar chimera_placeholder_4;
```

jslee02:
This is suppressed by #329, but it needs to be resolved correctly.

PR #329, opened 2020-05-05
Temporarily suppress dependent name types

jslee02:
Related issue: #328


test/examples/regression/issue328_dependent_name/issue328_dependent_name.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class TestIssue328(unittest.TestCase):
+
+    def _test_issue328(self, binding):
+        pass
+
+    def test_issue328(self):
+        self._test_issue328(py11)
+        self._test_issue328(boost)
```
psigen, line 14: In situations where we are just suppressing an issue instead of solving it, could we add an assertion that _should_ work if we didn't have the error (like being able to access this `::type`) and then comment it out in the test?

I am worried it will be hard to tell which issues we have solved and which ones we have just covered up, which will make it hard if we e.g. upgrade clang then some of this functionality is fixed.

jslee02:
> Before we suppress this, would it be possible to trace which string function is returning the unspecified template type instead of the fully-qualified template type?

I fully agree with you. Sorry, I should have provided more details.

The intention is actually resolving the issue in the end instead of just concealing it. Supporting the built-in type (#327) is in the path for example. This PR is just to suppress throwing the exception for now. I think we should print a warning saying that the API is ignored as it's not supported yet in the meantime.

psigen:
Ok, that makes sense.  Perhaps just add that placeholder unit test that shows the problem as per https://github.com/personalrobotics/chimera/pull/329/files#r419865135 and it should be fine to merge.

Issue #332, opened 2020-05-05
Generating empty docstring when Doxygen keyword encountered


jslee02:
```cpp
/// Returns the sum of two integer values  --> OK!
int add(int a, int b);

/// @brief Returns the sum of two integer values  --> Empty docstring generated
int add(int a, int b);
```

PR #333, opened 2020-05-05
Add regression test for #332

jslee02:


jslee02:
Yeah, I'm not sure what's the cause. Will take a look later :sweat_smile:

PR #339, opened 2020-05-11
Treat warnings as errors in CI builds

jslee02:


.ci/script.sh
```diff
@@ -13,9 +13,17 @@ mkdir build
 cd build
 
 if [ $BUILD_NAME = TRUSTY_GCC_DEBUG ]; then
-  cmake "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" "-DLLVM_DIR=${LLVM_DIR}" "-DCODECOV=ON" ..
+  cmake .. \
+    "-DCMAKE_BUILD_TYPE=${BUILD_TYPE}" \
+    "-DCHIMERA_TREAT_WARNINGS_AS_ERRORS=ON" \
+    "-DLLVM_DIR=${LLVM_DIR}" \
+    "-DCODECOV=ON"
```
psigen, line 20: Instead of this copy-paste, can we simply assign an intermediate variable like `USE_CODECOV` and have a single `cmake` command?

PR #340, opened 2020-05-11
Add more operator support

jslee02:


src/util.cpp
```diff
@@ -961,17 +961,42 @@ bool hasNonPublicParam(const CXXMethodDecl *decl)
...
+    {
+        getOperatorName(kind);
+        return true;
+    }
+    catch (std::invalid_argument)
+    {
+        return false;
+    }
+}
+
```
psigen, line 976: This seems silly.  If we have to do this we should just change the result of `getOperatorName` to `std::optional` or something, since this defeats the original purpose of the illegal argument indicating an invalid condition had been reached.

jslee02: I'm okay with the change. The purpose of this change was to keep syncing the supported types of `getOperatorName()` and `CXXRecord::methodsInternal()`.

psigen: As long as we think `getOperatorName()` is sufficiently lightweight, we can just make it explicitly get called, I think.    It's probably not a good idea to rely on throwing and catching an exception for a boolean check.

jslee02: Makes sense to me. Done!

psigen (approved):

